PR #15, opened 2017-03-30
Fix empty character classes

shvaikalesh:
I've fixed this way (in generator) to allow omitting `node.expressions = []` while traversing.


DmitrySoshnikov:
Perfect, thanks!

PR #16, opened 2017-03-30
Add strict mode

shvaikalesh:
Also prevents `quantifier` from leaking to global.


DmitrySoshnikov:
Sounds great, thanks!

PR #17, opened 2017-03-30
Make `flags` an alpha-sorted string

### LINKED ISSUE #11: Make flags an alpha-sorted string
Currently flags is an array, we should make it as an alphabetically sorted string, as in [RegExp.prototype.flags](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/flags).

```js
regexpTree.parse('/a/ig');
```

Should be:

```js
{
  "type": "RegExp",
  "body": {
    "type": "Char",
    "value": "a",
    "kind": "simple"
  },
  "flags": "gi"
}
```

shvaikalesh:
Closes #11.

### DIFF
--- a/src/parser/__tests__/parser-basic-test.js
+++ b/src/parser/__tests__/parser-basic-test.js
@@ -19,7 +19,7 @@ describe('basic', () => {
         value: 'a',
         kind: 'simple'
       },
-      flags: []
+      flags: '',
     });
   });
 
@@ -39,7 +39,7 @@ describe('basic', () => {
           kind: 'simple',
         }
       },
-      flags: [],
+      flags: '',
     });
   });
 
@@ -61,7 +61,7 @@ describe('basic', () => {
           }
         ],
       },
-      flags: [],
+      flags: '',
     });
   });
 
@@ -91,9 +91,7 @@ describe('basic', () => {
           }
         ]
       },
-      flags: [
-        'i'
-      ]
+      flags: 'i',
     });
   });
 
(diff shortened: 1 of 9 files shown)


src/parser/__tests__/parser-basic-test.js
```diff
@@ -19,7 +19,7 @@ describe('basic', () => {
         value: 'a',
         kind: 'simple'
       },
-      flags: []
+      flags: '',
```
DmitrySoshnikov: I'm wondering whether `null` would be better in case of no flags?

shvaikalesh: I think empty string is better.

Most common usages: check whether flag is enabled, remove flag, add flag. All three would require null checks (meh) and docs pointing that out. Furthermore, we are mimicking `RegExp#flags`, it returns an empty string.

shvaikalesh:
README is what I always forget, done.

src/parser/__tests__/parser-basic-test.js
DmitrySoshnikov: Yeah, good point, OK.

DmitrySoshnikov:
Cool, there is [one more](https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/parser/README.md) README in the parser directory.

shvaikalesh:
My bad, fixed that.

DmitrySoshnikov:
Nice! 👍

PR #18, opened 2017-03-30
Coerce to string in `parse`

### LINKED ISSUE #12: Add `fromRegExp` method
The `fromRegExp` method should accept an actual regular expression object, and return a parsed AST for it. This is a counterpart for the `toRegExp` method.

```js
const regexpTree = require('regexp-tree');

regexpTree.fromRegExp(/[a-z]+/i); // AST
regexpTree.fromRegExp(new RegExp('[a-z]+', 'i')); // AST
```

shvaikalesh:
Closes #12.

I would argue than instead of adding another method, `parse` should do `ToString`.

- It is simpler for users of API, makes first contact with module more pleasant ("Unexpected end of input" error is not very helpful).
- Less bloat.
- Will make `traverse` API simpler (`traverseFromRegExp` does not look right).
- Very JavaScripty and feels natural.

### DIFF
--- a/__tests__/integration-test.js
+++ b/__tests__/integration-test.js
@@ -74,4 +74,10 @@ describe('regexp-tree', () => {
     expect(re.test('Z')).toBe(true);
   });
 
+  it('calls `ToString` in `parse`', () => {
+    const reStr = '/m/m';
+    const ast = regexpTree.parse({toString: () => reStr});
+    expect(regexpTree.generate(ast)).toBe(reStr);
+  });
+
 });
--- a/index.js
+++ b/index.js
@@ -23,7 +23,7 @@ const regexpTree = {
    * @return Object AST
    */
   parse(regexp) {
-    return parser.parse(regexp);
+    return parser.parse(`${regexp}`);
   },
 
   /**


index.js
```diff
@@ -23,7 +23,7 @@ const regexpTree = {
    * @return Object AST
    */
   parse(regexp) {
-    return parser.parse(regexp);
+    return parser.parse(`${regexp}`);
```
DmitrySoshnikov: I think just `toString()` would be more explicit :)

__tests__/integration-test.js
```diff
@@ -74,4 +74,10 @@ describe('regexp-tree', () => {
     expect(re.test('Z')).toBe(true);
   });
 
+  it('calls `ToString` in `parse`', () => {
```
DmitrySoshnikov: Nit: `toString`

index.js
shvaikalesh: I avoid calling `toString` in JavaScript directly because it is just one step in `ToString` algorithm, along with `@@toPrimitive` and `valueOf`.

If I pass an object with valid `@@toPrimitive` or `valueOf` methods I expect them to be called. Also, `.toString()` on symbols will return description string, rather then throw like `ToString`. `RegExp` constructor uses `ToString`.

What about

```js
const ToString = v => `${v}`

// ...

return parser.parse(ToString(v))
```

DmitrySoshnikov: I see, OK, sounds good. We can leave w/o an extra method, as it's now.

DmitrySoshnikov:
Thanks for the fix!

PR #25, opened 2017-03-30
Minor tweaks

shvaikalesh:
Probably not the most useful PR, but still, worth a shot IMO.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "main": "index.js",
   "scripts": {
     "build": "./scripts/build",
-    "test": "./scripts/test",
+    "test": "jest",
     "prepublish": "npm test && npm run build"
   },
   "bin": {
--- a/package.json
+++ b/package.json
@@ -3,13 +3,9 @@
   "version": "0.0.19",
   "license": "MIT",
   "description": "Regular Expressions parser in JavaScript",
-  "repository": {
-    "type": "git",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -3,16 +3,12 @@
...
-    "url": "https://github.com/DmitrySoshnikov/regexp-tree.git"
-  },
+  "repository": "DmitrySoshnikov/regexp-tree",
   "homepage": "https://github.com/DmitrySoshnikov/regexp-tree",
   "bugs": "https://github.com/DmitrySoshnikov/regexp-tree/issues",
-  "main": "index.js",
   "scripts": {
     "build": "./scripts/build",
-    "test": "./scripts/test",
+    "test": "jest",
```
DmitrySoshnikov: Will this work if no global `jest` is installed? The purpose of the script was to use the local version of `jest` installed from `devDependencies`.

shvaikalesh: Yep, `npm` prepends ([docs](https://docs.npmjs.com/misc/scripts#path)) `node_modules/.bin` to `PATH` before running the code, so it does not require `jest` installed globally. Travis passed, and it does not have global `jest`.

DmitrySoshnikov: I see, didn't know about this. Sounds good!

DmitrySoshnikov:
OK, looks good overall!

PR #29, opened 2017-04-01
Support dotAll (\s) flag

### LINKED ISSUE #26: Parser: support "dotAll" \s flag
The new "dotAll" `\s` flag matches all symbols, including line terminators (https://github.com/mathiasbynens/es-regexp-dotall-flag). This proposal is at stage 3, so it's safe to add it. NOTE: `toRegExp` won't work with it, but having it in parser is fine.

shvaikalesh:
Closes #26.

### DIFF
--- a/src/parser/__tests__/parser-basic-test.js
+++ b/src/parser/__tests__/parser-basic-test.js
@@ -354,7 +354,7 @@ describe('basic', () => {
 
   it('valid not sorted flags', () => {
     // Not using `re` helper here because `RegExp.prototype.toString` sorts flags
-    expect(regexpTree.parse("/a/mgyiu")).toEqual({
+    expect(regexpTree.parse('/a/mgyiu')).toEqual({
       type: 'RegExp',
       body: {
         type: 'Char',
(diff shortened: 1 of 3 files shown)


src/parser/regexp.bnf
```diff
@@ -143,8 +143,10 @@ function Char(value, kind, loc) {
  */
 function checkFlags(flags) {
   const seen = new Set();
+  const valid = 'gimsuy';
```
DmitrySoshnikov: This probably can be moved outside `checkFlags` function as a const (right above the function).

src/parser/__tests__/parser-basic-test.js
```diff
@@ -365,4 +365,17 @@ describe('basic', () => {
     });
   });
 
+  it('dotAll (/s) flag', () => {
+    // Not using `re` helper here because /s flag is not yet implemented
```
DmitrySoshnikov: Nice! BTW, later we should be able to implement a transform to actually support it in regexes, transpiling `/./s` into `/[\0-\uFFFF]/`, and `/./su` into `/[\0-\u{10FFFF}]/u` as mentioned by @mathiasbynens.

PR #31, opened 2017-04-02
Add `symbol` on `Char` objects

### LINKED ISSUE #4: Show symbol for char-code nodes
For HEX, Octal, Decimal, and Unicode char codes we can show also an actual symbol corresponding to the code:

```
\u003B
```

Currently produces:

```js
{
  "type": "Char",
  "value": "\\u003B",
  "kind": "unicode"
}
```

We should add `symbol` property to the node:

```js
{
  "type": "Char",
  "value": "\\u003B",
  "kind": "unicode",
  "symbol": ";"
}
```

shvaikalesh:
Closes #4.

### DIFF
--- a/src/parser/generated/regexp-tree.js
+++ b/src/parser/generated/regexp-tree.js
@@ -758,7 +758,7 @@ function Char(value, kind, loc) {
 
   switch (kind) {
     case 'decimal': {
-      const code = parseInt(value.slice(1));
+      const code = Number(value.slice(1));
       symbol = String.fromCodePoint(code);
       break;
     }
--- a/src/parser/generated/regexp-tree.js
+++ b/src/parser/generated/regexp-tree.js
@@ -754,10 +754,33 @@ function getRange(text) {
  * Creates a character node.
  */
 function Char(value, kind, loc) {
+  let symbol;
+
+  switch (kind) {
+    case 'decimal': {
+      const code = parseInt(value.slice(1));
+      symbol = String.fromCodePoint(code);
+      break;
+    }
+    case 'oct': {
+      const code = parseInt(value.slice(1), 8);
+      symbol = String.fromCodePoint(code);
+      break;
+    }
+    case 'hex':
+    case 'unicode': {
+      const hex = value.slice(2).replace('{', '');
+      const code = parseInt(hex, 16);
+      symbol = String.fromCodePoint(code);
+      break;
+    }
+  }
+
   return Node({
     type: 'Char',
     value,
     kind,
+    symbol,
   }, loc);
 }
 
--- a/src/parser/__tests__/parser-basic-test.js
+++ b/src/parser/__tests__/parser-basic-test.js
@@ -263,7 +263,7 @@ describe('basic', () => {
         idx <= end;
         ++idx
       ) {
-        range.push(String.fromCharCode(idx));
+        range.push(String.fromCodePoint(idx));
       }
 
       return range;
--- a/src/parser/__tests__/parser-basic-test.js
+++ b/src/parser/__tests__/parser-basic-test.js
@@ -244,6 +244,7 @@ describe('basic', () => {
           {
             type: 'Char',
             value: '\\1',
+            symbol: String.fromCodePoint(1),
             kind: 'decimal'
           }
         ]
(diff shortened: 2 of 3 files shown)


src/parser/generated/regexp-tree.js
```diff
@@ -754,10 +754,33 @@ function getRange(text) {
  * Creates a character node.
  */
 function Char(value, kind, loc) {
+  let symbol;
+
+  switch (kind) {
+    case 'decimal': {
+      const code = parseInt(value.slice(1));
```
DmitrySoshnikov: We can use `Number(...)` coercion, it's faster, and does exactly coercion instead of parsing. Or otherwise, we need to put explicit `10` as radix for `parseInt`.

DmitrySoshnikov:
Cool stuff!

src/parser/generated/regexp-tree.js
shvaikalesh: nit: `Number` does parsing too, but a different one: the whole string should be parsed as number. The strictest for decimals is `parseInt(, 10)` (`10` is required to ban `0x`). `Number` allows `0[xob]` and floats.

shvaikalesh:
I am not quite happy with implementation, though.
I think it would be better to have `symbol` as a getter on `Char` class.

1. Creating objects without `new` is kinda awkward in modern JS.
2. With getter we can disallow setting it (simply by not providing setter; throws in strict mode). 
3. **neat**. `symbol` is computed from `value` and `kind`. Every time any of that props change, `symbol` will be in sync.
4. Compute `symbol` only when necessary.
5. By moving it to `.prototype`, we can remove it form parsers tests and test in different file.

DmitrySoshnikov:
Yeah, if it'd be "only for JS" we could for sure use actual AST node classes (with getters for needed properties, etc). Such AST nodes could even encapsulate methods for code generation, and `eval`ing for an interpreter.

However, I chose Mozilla Parser ASTs node format to be compatible with other tools (including AST Explorer, ast-traverse, etc), and also to be JSON-portable. I think it's fine for now with the symbol being as a prop.

shvaikalesh:
We can have both -- node classes and `toJSON` for JSON portability.

DmitrySoshnikov:
Yep, good idea, might reconsider it in the future if there are more, and bigger use-cases, when nodes will need to encapsulate more state. For now I think we can stick with the simpler format of plain objects (to avoid doing `toJSON` in other related tools).

shvaikalesh:
I am not quite excited about node manipulation methods in `handler`.

```
this.remove(node)
```

`node` should be removed from it's parent. But `this` is not it's parent. Hmmm.

```
this.replaceChild(a, b)
```

What is replaced with what? It is like old DOM API.

User provides empty object and we add methods / set `[[Prototype]]` on it.
User may set random props/methods on `handler` (like `this.dotAll` or some counters) and they may collide in future with node/flags manipulation.

DmitrySoshnikov:
Right, collisions might be (and actually will be, unless all "reserved" attached API hooks are specified in the docs) a problem. Alternatively, I'm thinking about passing `api` as a param to the handlers:

```js
regexpTree.transform('/[a-z]{1,}/', {
  onQuantifier(node, api) {
    ...
    api.removeNode(node);
  },
});
```

Although, now when I'm thinking more about it -- do we actually need an extra `transform` module? Since all operations can be covered already now in the `traverse`. With slightly "low-level" manipulations (you have to `delete` nodes, etc), but all the functionality (modify/replace/remove) is available. The purpose of an actual module I was thinking would be to provide exactly convenient methods (`remove`, `replaceWith`, etc). But now implementing them causing other issues, like where to attach them -- on `this`, on `node`, or not to attach at all? This seems will be duplicating of the functionality, and may cause questions like "what to use for transformation -- `transform` or `traverse`?".

If the `transform` won't be a separate module, it can be just a small wrapper on top of "parse-traverse-generate" as it initially was planned (to be able to pass an actual regexp instead of an AST, and don't call generator manually).

DmitrySoshnikov:
And for convenient methods it might be then `transformUtil` light module, which would accept explicit `node`, `parent`, `prop`, etc. if needed.

PR #33, opened 2017-04-03
Throw on invalid unicode escapes

### LINKED ISSUE #13: It should be SyntaxError if Unicode code point is greater than 0x10FFFF
This should be an invalid regexp:

```js
/\u{0x11FFFF}/u
```

The max value which should be checked is `0x10FFFF`.

shvaikalesh:
Closes #13.

### DIFF
--- a/src/parser/generated/regexp-tree.js
+++ b/src/parser/generated/regexp-tree.js
@@ -771,6 +771,10 @@ function Char(value, kind, loc) {
     case 'unicode': {
       const hex = value.slice(2).replace('{', '');
       const code = parseInt(hex, 16);
+      if (code > 0x10ffff) {
+        throw new SyntaxError(`Bad character escape sequence: ${value}`);
+      }
+
       symbol = String.fromCodePoint(code);
       break;
     }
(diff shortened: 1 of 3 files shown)


src/parser/generated/regexp-tree.js
```diff
@@ -771,6 +771,10 @@ function Char(value, kind, loc) {
     case 'unicode': {
       const hex = value.slice(2).replace('{', '');
       const code = parseInt(hex, 16);
+      if (code > 0x10ffff) {
+        throw new SyntaxError(`Bad character escape sequence: ${value}`);
```
DmitrySoshnikov: It'd be good also to put location information (if `loc` is not null).

```js
// top-level, can be reused in other manual error messages.
function getLocString(loc) {
  return loc ? ` at ${loc.start}:${loc.end}` : '';
}

...

throw new SyntaxError(
  `Bad character escape sequence: ${value}${getLocString(loc)}.`);
);
```

DmitrySoshnikov: Sorry, should be ` at 1:${loc.start}` (since we have only 1-liner regexes).

Issue #34, opened 2017-04-03
Development: Add Flow types for static types safety


DmitrySoshnikov:
Currently the code base is not statically type checked, we can use [Flow](https://flow.org/) type checker to validate it.

PR #39, opened 2017-04-05
remove some commas that generated errors

hg42:
fixes some errors with trailing commas in function parameters when not having ES2017
see [Trailing commas - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas)

### DIFF
--- a/src/traverse/__tests__/node-path-test.js
+++ b/src/traverse/__tests__/node-path-test.js
@@ -36,7 +36,7 @@ describe('NodePath', () => {
       ast.body.expressions[1],
       new NodePath(ast.body),
       'expressions',
-      1,
+      1
     );
 
     bCharPath.remove();
@@ -67,7 +67,7 @@ describe('NodePath', () => {
       ast.body.expressions[1],
       new NodePath(ast.body),
       'expressions',
-      1,
+      1
     );
 
     const cNode = {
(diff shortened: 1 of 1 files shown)


DmitrySoshnikov:
Ah, good catch; thanks!

mathiasbynens:
Why was this not caught by tests?

DmitrySoshnikov:
Jest's preprocessor transformed it to ES5, using standard Babel presets for ES2015, and other stages.

The code base is mostly ES6+ (with latest Node6+), and doesn't use a transpiler. Thought probably should add an explicit build step, and transpile. We'll have to anyways for #34.

DmitrySoshnikov:
Though I'm confused now how @hg42 caught those errors? Was an older version of Node? If Jest transpiles it anyways, shouldn't be an issue.

hg42:
I am fairly new to the javascript world and still learning on a very steep curve.
I hope I am doing it the right way.

I simply used `npm test`.

I am using `node v6.9.2` from debian/experimental,
and I have `jest-cli@19.0.2` in regexp-tree (jest is also globally installed with the same version).

Which node version do you use?

hg42:
From what I know, commas are allowed in [] and {} in ES2015, but commas in function parameters were added in ES2017. So babel with ES2015 shouldn't be sufficient, right?

PR #40, opened 2017-04-05
add "None" type to generator which generates no code

hg42:
this allows transformations to simply set the type of a node to `"None"` instead of removing it, when it's sufficient to keep the dead nodes until `generate()`

### DIFF
--- a/src/generator/index.js
+++ b/src/generator/index.js
@@ -20,6 +20,10 @@ const generator = {
     return `/${gen(node.body)}/${node.flags}`;
   },
 
+  None(node) {
+    return '';
+  },
+
   Alternative(node) {
     return (node.expressions || [])
       .map(gen)


DmitrySoshnikov:
This is a good workaround, though might be fragile, and is really a workaround/hack. We introduce "fake" (virtual) node type, which is not related to `RegExp` (which will require documenting it, etc). Also, some classes, e.g. `TransformResult` expose the AST, via `getAST()` method, which will return to callers this `"None"` type, so they will have to know about it.

I think for now we should just keep it as a workaround, and do this in callers as in your example, and look into fixing the `NodePath`'s `remove` method, etc.

DmitrySoshnikov:
I'll try to take a look at `remove` working tonight (with removing dependency on `ast-traverse`, and adjusting `index` when needed), and also add `getNextSibling` which will return `NodePath` instead.

DmitrySoshnikov:
OK, the index issue after removal, and `getNext/PreviousSibling` should be fixed in https://github.com/DmitrySoshnikov/regexp-tree/commit/9859d991bd20f98f863ca6559dbd932a152a1d69 (and in `0.0.28`).

hg42:
ok, agreed, I'll close it

PR #43, opened 2017-04-06
package.json: Add `engines`

mathiasbynens:
Explicitly state that this projects supports Node.js v6+ only.

https://github.com/DmitrySoshnikov/regexp-tree/pull/39#issuecomment-292106505


DmitrySoshnikov:
Cool, thanks!

PR #47, opened 2017-04-08
add getParent() to NodePath

hg42:
we need this
* to navigate to parent without knowing implementation details (NodePath, node.parent etc.)
* to have a consistent API

### DIFF
--- a/src/traverse/__tests__/node-path-test.js
+++ b/src/traverse/__tests__/node-path-test.js
@@ -287,21 +287,19 @@ describe('NodePath', () => {
   it('getParent/getChild', () => {
     const ast = parser.parse('/a(bc)d/');
 
-    //console.log(JSON.stringify(ast, null, '  '));
-
-    bodyPath = NodePath.getForNode(ast.body);
-    groupPath = bodyPath.getChild(1);
+    const bodyPath = NodePath.getForNode(ast.body);
+    const groupPath = bodyPath.getChild(1);
 
     expect(groupPath.node.type).toBe("Group");
     expect(groupPath.getParent()).toBe(bodyPath);
 
-    alterPath = groupPath.getChild();
+    const alterPath = groupPath.getChild();
 
     expect(alterPath.node.type).toBe("Alternative");
     expect(alterPath.getParent()).toBe(groupPath);
 
-    bCharPath = alterPath.getChild(0);
-    cCharPath = alterPath.getChild(1);
+    const bCharPath = alterPath.getChild(0);
+    const cCharPath = alterPath.getChild(1);
 
     expect(bCharPath.getParent()).toBe(alterPath);
     expect(cCharPath.getParent()).toBe(alterPath);
--- a/src/traverse/__tests__/node-path-test.js
+++ b/src/traverse/__tests__/node-path-test.js
@@ -281,6 +281,44 @@ describe('NodePath', () => {
     );
 
     expect(aCharPath.getNextSibling()).toBe(bCharPath);
+
+  });
+
+  it('getParent/getChild', () => {
+    const ast = parser.parse('/a(bc)d/');
+
+    //console.log(JSON.stringify(ast, null, '  '));
+
+    bodyPath = NodePath.getForNode(ast.body);
+    groupPath = bodyPath.getChild(1);
+
+    expect(groupPath.node.type).toBe("Group");
+    expect(groupPath.getParent()).toBe(bodyPath);
+
+    alterPath = groupPath.getChild();
+
+    expect(alterPath.node.type).toBe("Alternative");
+    expect(alterPath.getParent()).toBe(groupPath);
+
+    bCharPath = alterPath.getChild(0);
+    cCharPath = alterPath.getChild(1);
+
+    expect(bCharPath.getParent()).toBe(alterPath);
+    expect(cCharPath.getParent()).toBe(alterPath);
+
+    expect(bCharPath.getParent().getParent()).toBe(groupPath);
+    expect(cCharPath.getParent().getParent()).toBe(groupPath);
+
+    expect(groupPath.getChild()).toBe(alterPath);
+    expect(groupPath.getChild(0)).toBe(alterPath);
+    expect(groupPath.getChild(1)).toBe(null);
+
+    expect(groupPath.getChild().getChild(0)).toBe(bCharPath);
+    expect(groupPath.getChild().getChild(1)).toBe(cCharPath);
+
+    expect(groupPath.getChild()).toBe(alterPath);
+    expect(groupPath.getChild(0)).toBe(alterPath);
+    expect(groupPath.getChild(1)).toBe(null);
   });
 
 });
--- a/src/traverse/node-path.js
+++ b/src/traverse/node-path.js
@@ -113,6 +113,16 @@ class NodePath {
   }
 
   /**
+   * Returns parent.
+   */
+  getParent() {
+    if (!this.parent) {
+      return null;
+    }
+    return NodePath.getForNode(this.parent);
+  }
+
+  /**
    * Returns previous sibling.
    */
   getPreviousSibling() {
--- a/src/traverse/node-path.js
+++ b/src/traverse/node-path.js
@@ -123,6 +123,23 @@ class NodePath {
   }
 
   /**
+   * Returns nth child.
+   */
+  getChild(n = 0) {
+    if (this.node.expressions) {
+      return NodePath.getForNode(
+        this.node.expressions[n],
+        this,
+        "expressions",
+        n
+      );
+    } else if (this.node.expression && n == 0) {
+      return NodePath.getForNode(this.node.expression, this, "expression");
+    }
+    return null;
+  }
+
+  /**
    * Returns previous sibling.
    */
   getPreviousSibling() {
--- a/src/traverse/node-path.js
+++ b/src/traverse/node-path.js
@@ -116,10 +116,7 @@ class NodePath {
    * Returns parent.
    */
   getParent() {
-    if (!this.parent) {
-      return null;
-    }
-    return NodePath.getForNode(this.parent);
+    return this.parentPath;
   }
 
   /**
--- a/src/traverse/README.md
+++ b/src/traverse/README.md
@@ -52,3 +52,5 @@ Methods of the `NodePath`:
 * `update(nodeProps: Object)` - updates a node inline;
 * `getPreviousSibling()` - returns previous sibling path;
 * `getNextSibling()` - returns next sibling path;
+* `getChild(n: Number = 0)` - return `n`th child path;
+* `getParent()` - return parent path, for consistency, returns parentPath;


hg42:
does not work as expected, had to write a better test first.
getParent() usually works one level but not two.

I reopened this to discuss if it's useful for one level...because I used it already.
But  the API induces to believe it could be used twice or more.
What do you think.

I had expected a NodePath would be found magically. This could be done but would need traversal to find the node in the tree. For that the tree would have to be stored somewhere.

I also have a getChild(n), which I wanted to add to this PR, but my tests currently do not work.

DmitrySoshnikov:
@hg42, I'm currently on a small vacation, and out of laptop, writing from phone. Will take a look once will come back after weekend on Monday.

DmitrySoshnikov:
Took a brief look, yeah, single parent doesn't sound right. Also, the `NodePath` abstraction was created exactly to represent a _path_, i.e. to be able to access a chain of _all parent nodes_, by accessing the `parentPath` property on a path.

So this would give you access to any parent level:

```js
regexpTree.transform('/(a[b])/', {
  Char(path) {
    if (path.node.value === 'b') {
      console.log(
        path.node.type, // Char (itself)
        path.parentPath.node.type, // CharacterClass
        path.parentPath.parentPath.node.type, // Alternative
        path.parentPath.parentPath.parentPath.node.type // Group
        // ... etc
      );
    }
  }
};
```

For convenience `NodePath` has also `parent` property, which is a direct parent node, i.e.:

```js
path.parent === path.parentPath.node; // true
```

Re: `getChild` - it also seems to me "too specific" that it works only with `expression`/`expressions`. And it's trying to apply model of the generic `childNodes`. As we've been discussing, this tries to follow Mozilla Parser API (or newer version of it now -- [ESTree](https://github.com/estree/estree/blob/master/es5.md)) format. It assumes that callers _should know_ the structure of the nodes they deal with. I.e. they should know that a specific property is called e.g. `expressions` -- from which they can access needed child as `expressions[n]`. But this is only for nodes. If we want to get child paths, maybe it could be as:

```js
path.getChildPath('expressions', 3);
```

Specifying always the `'expressions'` might at first glance look like an annoying boiler plate (and I would agree it is :)), however it would be more explicit, and still following perception that it's not a "DOM-like generic childNodes" format, but underlying that it's ESTree-based format, and that callers should know which properties they try to access (passing `'expressions'` as a string might be error-prone to typos though).

...

Actually, while typing I though probably a generic `path.forNode` could be even more useful:

```js
path.forNode(node.expressions[3]);
```

In this case no typos with passing strings for property names, and a path will be returned. Thoughts?

DmitrySoshnikov:
Actually (and unfortunately), simple `path.forNode(node.expressions[3]);` seems won't work, since we need to pass full context (parent path, prop, index) in case if a path doesn't exist yet in registry, and we need to create one.

hg42:
From what I know now, it's like that:
* NodePaths are created when needed
* inside `traverse()` you already have all NodePaths towards the root of the AST already created (I think), but not those for children
* to create a NodePath for a child in an array you have to know several details (who is the parent, which property is used for the array and the index)
* if you don't use traverse (as I do at some place working directly with the bottom level of the expression), you don't have NodePaths already created.

That's why I felt a need to encapsulate theses operations.

One part of the problem seems to be that NodePath and Node is separate, but to work with a NodePath you need to know details only a Node can know.

You were able to implement `getSibling()` without knowing the array property only because fortunately it's stored in your current node (only the same because it's a sibling).

I see these solutions:
* the children are stored in a common way, so you don't need details of the Node
* nodes are is an objects and can give you the property name you need (`getChildrenPropertyName()`?)
* somewhere is a registry holding the relation of node type to children property. The relation depends on the syntax elements, so it should probably be defined there or generated from it

DmitrySoshnikov:
Yeah, we could solve it simpler with 2-pass approach: on first pass all nodes are wrapped into `NodePath`, the second actually handles the paths. Currently we create a `NodePath` during traversal (or when access next/previous sibling), so yeah -- previous paths are guaranteed to exists, and the nodes which aren't handled yet -- are not guaranteed (but can in case of collection nodes, due to next/previous sibling path creation).

That's said, abstracting away the `expression(s)` might be hiding the details (which aren't in fact details), though now I think it might be a useful solution.

So yeah, let's take the `getChild(n = 0)` with your implementation with defaulting to `exporession(s)` property. Probably we'll be able to provide a second (optional) parameter for property name.

And the `getParent()` seems is unnecessary as discussed.

DmitrySoshnikov:
Also, we'll need to update [README](https://github.com/DmitrySoshnikov/regexp-tree/tree/master/src/traverse#nodepath-class) for the `getChild`.

DmitrySoshnikov:
FYI: in https://github.com/DmitrySoshnikov/regexp-tree/commit/3e537a42923407b0c99144bcf61817a35755b3ac I also added `insertChildAt`, which will correlate with `getChild`, and also default property name to `expressions`.

hg42:
(seems I forgot to post this yesterday)

yesterday I wrote some reordering operations on the tree for the group position problem.

test input (a bit wild :-) using x-flag):
`/   (bcd) ( [xey] .*  \n  (g.* (?<I>i) ) \n jk (?<L>l) ) (?:mn)? (opq)  r s t /`

wrap all the non-grouped parts in groups (to cummulate the position in the match result):
`/((?:b)(?:c)(?:d))((?:[xey])(?:.*)((?:g)(?:.*)(?<I>i))(?:j)(?:k)(?<L>l))(?:(?:(?:m)(?:n))?)((?:o)(?:p)(?:q))(?:r)(?:s)(?:t)/`

then optimize to (combine adjacent wrapped groups in alternative):
`/(bcd)((?:[xey].*)((?:g.*)(?<I>i))(?:jk)(?<L>l))(?:(?:mn)?)(opq)(?:rst)/`

final javascript regexp (all capturing groups):
`/(bcd)(([xey].*)((g.*)(i))(jk)(l))((mn)?)(opq)(rst)/`

wrapping was simple.

but then I tried to combine the groups into an Alternative in a Group:
`((?:b)(?:c)(?:d))` -> `((bcd))`
so c should be appended to b, and d to bc.

I wanted to add the expression of a group to the previous if it meets some condition (being a wrapped group, which is marked in the pass before by node.wrapped=true), eventually inserting an alternative in between when not existent. Simplified pseudo code:
```
Group:
  // doing some other things with the Group
  previous = node.getPreviousSibling()
  if previous is Group and wrapped
    if previous.expression is not Alternative
     previous.expression = Alternative([previous.expression])
    previous.expression.push(node)
    node.remove()    // should this eventually read removeFromParent()?
```
That's similar to how I think and would ideally like to write it down.

Creating the NodePaths was driving me crazy. The code was three times as big because of all these long NodePath lines.
It always ended in a path.node being null. So it seems I had a wrong path.
But may be I had another problem and just didn't see it (it was late in the night).

I also tried several ways to set a node. E.g. path.replace(node).

At the end I very much simplified this by using a second pass and removing all the NodePath lines.

How would you write this transformation?
Which NodePaths should be created for the new nodes?

Today I think, I should have had a look on the parent/parentPath members.

hg42:
insertChildAt looks good.

You could eventually consider to change this:
```js
if (index <= NodePath.getTraversingIndex()) {
  NodePath.updateTraversingIndex(+1);
}
```
to
```js
NodePath.updateTraversingIndex(+1, index);   // may be parameter "at"
```

hg42:
I think, if keeping the NodePaths, the set of methods on NodePaths should be more complete, so it hides paths from everything you do with nodes.
Currently, I am not sure what exactly should be added.

In general this would convert any `node.operation(...)` to a `path.operation(...)`.
This could be limited to certain operations, if the rules can be clearly described.
E.g, it's no problem to say `path.node.type`, but the rules should be simple and unambiguous.

`insertChildAt()` is a step in this direction.
Perhaps `appendChild()` could be another (could be insert with a certain parameter, but null or no parameter at all would be seen as 0, may be -1 for end?).

I think, `path.replace(node)` should also manage the NodePath by default, because it's a method of NodePath and it has all necessary information. A second parameter could suppress this, but I doubt anyone will need this.

hg42:
above pseudo code looks like this (first traverse) when using a second pass and not creating NodePaths:
```js
      RegExpTree.traverse(ast, {

        Group: function(path) {
          if(!path.node.wrapped)
            return;
          var prevSibling = path.getPreviousSibling();
          if( prevSibling &&
              prevSibling.node.type == "Group" &&
              prevSibling.node.wrapped) {
            if(prevSibling.node.expression.type != "Alternative") {
              prevSibling.node.expression = {
                type: "Alternative",
                expressions: [ prevSibling.node.expression ]
              }
            }
            prevSibling.node.expression.expressions.push(path.node.expression);
            path.remove();
          }
        }
      });
```
and the second pass removes groups in groups (the inner group is always a wrapper and can be removed)
```
      RegExpTree.traverse(ast, {

        Alternative: function(path) {
          if( path.node.expressions.length == 1 &&
              path.node.expressions[0].type == "Group"
            ) {
            path.replace(path.node.expressions[0].expression);
          }
        }
      });
```
Typically, you can work in two directions:
* outside-in: in the collection: reorder the children
* inside-out: in the child: reorder yourself
sometimes it helps to choose the other direction, if it becomes difficult.
May be that's the reason why my current variant was easier.
Unfortunately I have simply thrown away the first variant, so I don't know what was going on.

In the code above may still be several issues (with NodePaths).
I think they don't show up, because the following traverse() passes don't need path navigation.
The first is
```js
              prevSibling.node.expression = {
                type: "Alternative",
                expressions: [ prevSibling.node.expression ]
              }
```
where the (existing) path belonging to `prevSibling.node.expression` will stay on the old parent.
This kind of insertion of a layer should occur frequently because groups have only one child.
`(a)` and `(ab)` are very different in structure. This could eventually be straightened. Either by having a group always have a collection as expression, even if it has only one element. Or let a group to be an Alternative by definition. That's another case of "theoretical consistent" against "practical more consistent". Not sure, if I would want to do this. This could also be done on another level:
E.g. `group.append(node)` could automatically insert an Alternative, when a second child is appended.

The second is
```js
            prevSibling.node.expression.expressions.push(path.node.expression);
```
where the the path of `path.node.expression` will have a wrong parent.
This could be fixed automatically by an `appendChild()` if it creates and checks paths.

Also in the second `traverse()`:
```js
            path.replace(path.node.expressions[0].expression);
```
will leave the wrong parent in the path of `path.node.expressions[0].expression`.
This could be fixed by replace automatically.

(btw. I am traversing eight times now)

DmitrySoshnikov:
@hg42, thanks, all valid points on inconvenience of handling some very nested structures. I'll take a look a bit later (and on the use-case with groups, and alternatives) -- currently got a bit busy with main work (this is mostly my side project).

At the mean time, can we merge just `getChild` for now (I'll need in one transform)? The `getParent` is still obsolete, since one of the  `NodePath`'s reasons was exactly to provide `path.parentPath.parentPath`, and abstracting it further might be unnecessary.

Or if you'd still like to work with exactly `getParent`, I think we can change it to:

```js
getParent() {
  return this.parentPath;
}
```

So you'll be able to access:

```
path.getParent().getParent();
```

hg42:
yes, I thought of accessing parentPath, too.
I will change it...I'll also look at the README

src/traverse/__tests__/node-path-test.js
```diff
@@ -281,6 +281,44 @@ describe('NodePath', () => {
...
     expect(aCharPath.getNextSibling()).toBe(bCharPath);
+
+  });
+
+  it('getParent/getChild', () => {
+    const ast = parser.parse('/a(bc)d/');
+
+    //console.log(JSON.stringify(ast, null, '  '));
+
+    bodyPath = NodePath.getForNode(ast.body);
```
DmitrySoshnikov: Use `const`, or `let` (if a var is supposed to be changed), when define variables. Assigning without const/let creates global variable (basically it escapes to global, and pollutes it).

```js
const bodyPath = ...
```

Ditto for other local vars created.

src/traverse/node-path.js
```diff
@@ -113,6 +113,33 @@ class NodePath {
   }
 
   /**
+   * Returns parent.
+   */
+  getParent() {
```
DmitrySoshnikov: That's said, this seems obsolete, since doesn't abstract anything. If you'd like to have `getParent` for consistency with `getChild` (with might be a good reason), let's change it to:

```js
/**
 * Returns parent path.
 */
getParent() {
  return this.parentPath;
}
```

src/traverse/__tests__/node-path-test.js
hg42: yeah, sorry, I got used to livescript (and coffeescript in Atom)

DmitrySoshnikov:
OK, perfect, thanks!

(closing the PR, but we can continue discussion about other issue either on this thread, or to open another issues)

DmitrySoshnikov:
BTW, `getChild` actually turned out to be pretty useful :) I used it [in Repetition](https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/optimizer/transforms/char-star-to-plus-transform.js#L28) node, and [in Class Range](https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/optimizer/transforms/char-class-to-meta-transform.js#L35). We can probably add `body` to `getChild` as well, to handle top-level `RegExp` node.

PR #61, opened 2017-04-15
fix for parent and parentPath not in sync and index != null for single child

### LINKED ISSUE #60: reordering `a(b)` to `b(a)` does not set parentPath correctly (parent not in sync with parentPath)
I have a test case, where the parents get out of sync.

I think there is something wrong or missing in my latest `setChild` code or perhaps in `getForNode`.

Originally I used `replace()` in the test case and blamed it for the error.
But now I have rewritten it using `setChild`. I generally think `replace()` shouldn't be used for such reorders. But now I have a very similar error.

I already tried to find the bug, but I didn't succeed, yet.
My debugger is non-functional currently, so I only try to use logging and good conclusions.

Perhaps someone else wants to try?

But of course I'm not out of options (I guess I never will), so I'll try more myself.
I just wanted to let everyone know, that there is a problem.

This is the test case:
```js
  it('replaces with already used node with another parent', () => {
    const ast = parser.parse('/a(b)/');

    const bodyPath = new NodePath(ast.body);
    const aCharPath = bodyPath.getChild(0);
    const groupPath = bodyPath.getChild(1);
    const bCharPath = groupPath.getChild();
    const aNode = aCharPath.node;
    const bNode = bCharPath.node;

    bodyPath.node.mark = "body";
    groupPath.node.mark = "group";

    // swap a and b
    groupPath.setChild(aNode);
    bodyPath.setChild(bNode, 0);

    // body and group still at same position
    expect(bodyPath.node.mark).toEqual('body')
    expect(groupPath.node.mark).toEqual('group')
    expect(bodyPath.getChild(1).node.mark).toEqual('group')
    // second is Char 'a' in the Group
    expect(bodyPath.getChild(1).getChild().node.type).toEqual('Char')
    expect(bodyPath.getChild(1).getChild().node.value).toEqual('a')
    // now the parent of 'a' is the Group
    expect(bodyPath.getChild(1).getChild().parent.mark).toEqual('group')
    
    // ...
  });
```
the last `expect` fails like this:
```
    Expected value to equal:
      "group"
    Received:
      "body"
```

If you add this (instead of the last `expect`):
```js
expect(bodyPath.getChild(1).getChild().parentPath.node)
 .toBe(bodyPath.getChild(1).getChild().parent)
```
you get this error:
```
    - Expected
    + Received

    Object {
    -  "expressions": Array [
    -    Object {
    -      "kind": "simple",
    -      "symbol": undefined,
    -      "type": "Char",
    -      "value": "b",
    -    },
    -    Object {
    -      "capturing": true,
    -      "expression": Object {
    -        "kind": "simple",
    -        "symbol": undefined,
    -        "type": "Char",
    -        "value": "a",
    -      },
    -      "mark": "group",
    -      "number": 1,
    -      "type": "Group",
    -    },
    -  ],
    -  "mark": "body",
    -  "type": "Alternative",
    +  "capturing": true,
    +  "expression": Object {
    +    "kind": "simple",
    +    "symbol": undefined,
    +    "type": "Char",
    +    "value": "a",
    +  },
    +  "mark": "group",
    +  "number": 1,
    +  "type": "Group",
     }
```
so the node in the `parentPath` of 'a' is the body (which is wrong) and the `parent` is the group (which is correct). Of course they should be identical (=group).

@DmitrySoshnikov you added the same `NodePath.getForNode` to `replace()` like I did.
I still think, this is correct.

EDIT: I removed some lines about getForNode and index parameter, because that's not ready to be discussed

hg42:
* fixes #60
* fixes index != null for single child

### DIFF
--- a/src/traverse/__tests__/node-path-test.js
+++ b/src/traverse/__tests__/node-path-test.js
@@ -335,47 +335,47 @@ describe('NodePath', () => {
     bodyPath.setChild(bNode, 0);
 
     // 'a' and 'b' should be swapped now
-    expect(generator.generate(ast)).toEqual('/b(a)/');
+    expect(generator.generate(ast)).toBe('/b(a)/');
 
     // body and group still ok
-    expect(bodyPath.node.mark).toEqual('body')
-    expect(groupPath.node.mark).toEqual('group')
-    expect(bodyPath.getChild(1).node.mark).toEqual('group')
+    expect(bodyPath.node.mark).toBe('body')
+    expect(groupPath.node.mark).toBe('group')
+    expect(bodyPath.getChild(1).node.mark).toBe('group')
 
     // first is now Char 'b' in body
-    expect(bodyPath.getChild(0).node.type).toEqual('Char')
-    expect(bodyPath.getChild(0).node.value).toEqual('b')
+    expect(bodyPath.getChild(0).node.type).toBe('Char')
+    expect(bodyPath.getChild(0).node.value).toBe('b')
     // parentPath should point to the same node as parent
     expect(bodyPath.getChild(0).parentPath.node).toBe(bodyPath.getChild(0).parent)
     // now the parent of 'b' is the body
-    expect(bodyPath.getChild(0).parent.mark).toEqual('body')
-    expect(bodyPath.getChild(0).parent.type).toEqual('Alternative')
+    expect(bodyPath.getChild(0).parent.mark).toBe('body')
+    expect(bodyPath.getChild(0).parent.type).toBe('Alternative')
     // Group has children property 'expressions'
-    expect(bodyPath.getChild(0).property).toEqual('expressions')
+    expect(bodyPath.getChild(0).property).toBe('expressions')
     // and an index of 0
-    expect(bodyPath.getChild(0).index).toEqual(0)
+    expect(bodyPath.getChild(0).index).toBe(0)
 
     // second is Char 'a' in the Group
-    expect(bodyPath.getChild(1).getChild().node.type).toEqual('Char')
-    expect(bodyPath.getChild(1).getChild().node.value).toEqual('a')
+    expect(bodyPath.getChild(1).getChild().node.type).toBe('Char')
+    expect(bodyPath.getChild(1).getChild().node.value).toBe('a')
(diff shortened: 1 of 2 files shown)


src/traverse/__tests__/node-path-test.js
```diff
@@ -368,7 +429,7 @@ describe('NodePath', () => {
     expect(dPath.parentPath).toBe(alterPath);
     expect(alterPath.parentPath).toBe(groupPath);
 
-    expect(generator.generate(ast)).toBe('/a(cd)/');
+    expect(generator.generate(ast)).toEqual('/a(cd)/');
```
DmitrySoshnikov: nit for the future: `toBe` is usually fine with primitive values (like strings)

Usually, `toBe` is used to check the "identity", and `toEqual` to check a complex "shape" of an object.

```js
const foo = {x: 1};

// ok, same memory location
expect(foo).toBe(foo);

// fails, since two different objects in memory
expect(foo).toBe({x: 1});

 // ok, shapes are the same
expect(foo).toEqual({x: 1});
```

Since strings are primitive data type, they are always equal with `toBe` (and technically with `toEqual`).

src/traverse/__tests__/node-path-test.js
```diff
@@ -317,6 +317,67 @@ describe('NodePath', () => {
...
+
+    // the locally stored NodePaths should have new parents
+    expect(bCharPath.parentPath).toBe(bodyPath);
+    expect(aCharPath.parentPath).toBe(groupPath);
+    // paths from the registry should be the same
+    const bCharPath2 = NodePath.getForNode(bNode);
+    expect(bCharPath2).toBe(bCharPath);
+    const aCharPath2 = NodePath.getForNode(aNode);
+    expect(aCharPath2).toBe(aCharPath);
+  });
```
DmitrySoshnikov: Looks good, thanks 👍

src/traverse/__tests__/node-path-test.js
hg42: I noticed it already and added a commit :-)

DmitrySoshnikov: Ah, perfect, thanks!

PR #68, opened 2017-04-26
Fixing character group parsing with punctuation symbols

silentroach:
`/ ([.|,])/g` is now valid regexp

It is magic a little for me, so I just hope I will not break something with this fix :)


DmitrySoshnikov:
Doh! Really good catch; thanks, @silentroach!

PR #71, opened 2017-05-05
convert single chars group disjunction to chargroup

silentroach:
(a|b|c) -> ([abc])
(?:a|b|c) -> [abc]

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+
+[*.js]
+indent_style = space
+indent_size = 2
--- a/src/optimizer/transforms/__tests__/group-single-chars-to-char-class-test.js
+++ b/src/optimizer/transforms/__tests__/group-single-chars-to-char-class-test.js
@@ -1,0 +1,48 @@
+/**
+ * The MIT License (MIT)
+ * Copyright (c) 2017-present Dmitry Soshnikov <dmitry.soshnikov@gmail.com>
+ */
+
+'use strict';
+
+const {transform} = require('../../../transform');
+const singleCharsGroupToCharClass = require('../group-single-chars-to-char-class');
+
+describe('(a|b|c) -> ([abc])', () => {
+
+  it('replaces single chars group to char class', () => {
+    const re = transform(/(a|b|c)/, [
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe('/([abc])/');
+  });
+
+  it('works with metachars too', () => {
+    const re = transform(/(\d|\n)/, [
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe(/([\d\n])/.toString());
+  });
+
+  it('merges duplicates in group', () => {
+    const re = transform(/(a|b|a|b)/, [
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe('/([ab])/');
+  });
+
+  it('removes non-capturing group', () => {
+    const re = transform(/(?:a|b|c)/, [
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe('/[abc]/');
+  });
+
+  it('have no effect on multichar values', () => {
+    const re = transform(/(a|b|no)/, [
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe('/(a|b|no)/');
+  });
+
+});
--- a/src/optimizer/transforms/group-single-chars-to-char-class.js
+++ b/src/optimizer/transforms/group-single-chars-to-char-class.js
@@ -1,0 +1,57 @@
+/**
+ * The MIT License (MIT)
+ * Copyright (c) 2017-present Dmitry Soshnikov <dmitry.soshnikov@gmail.com>
+ */
+
+'use strict';
+
+/**
+ * A regexp-tree plugin to replace single char group disjunction to char group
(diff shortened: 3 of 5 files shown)


src/optimizer/transforms/__tests__/group-single-chars-to-char-class-test.js
```diff
@@ -0,0 +1,48 @@
...
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe(/([\d\n])/.toString());
+  });
+
+  it('merges duplicates in group', () => {
+    const re = transform(/(a|b|a|b)/, [
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe('/([ab])/');
```
DmitrySoshnikov: Awesome 👍

src/optimizer/transforms/group-single-chars-to-char-class.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * A regexp-tree plugin to replace single char group disjunction to char group
+ *
+ * (a|b|c) -> ([abc])
+ * (?:a|b|c) -> [abc]
+ */
+module.exports = {
+  Group(path) {
+    const { node } = path;
```
DmitrySoshnikov: super nit on style: no spaces `const {node} = path;`

src/optimizer/transforms/group-single-chars-to-char-class.js
```diff
@@ -0,0 +1,54 @@
...
+
+    const charset = new Map();
+
+    if (!processExpressionCharset(node.expression, charset)
+      || !charset.size
```
DmitrySoshnikov: nit: `||` on previous line

```js
if (!processExpressionCharset(node.expression, charset) ||
    !charset.size) {
  ...
}
```

Or:

```js
if (
  !processExpressionCharset(node.expression, charset) ||
  !charset.size
) {
  ...
}
```

src/optimizer/transforms/group-single-chars-to-char-class.js
```diff
@@ -0,0 +1,54 @@
...
+
+    const characterClass = {
+      type: 'CharacterClass',
+      expressions: Array.from(charset.values())
+    };
+
+    if (node.capturing) {
+      node.expression = characterClass;
+    }
+    else path.replace(characterClass);
```
DmitrySoshnikov: nit: explicit block even for one statement

```js
if (node.capturing) {
  node.expression = characterClass;
} else {
  path.replace(characterClass);
}
```

src/optimizer/transforms/group-single-chars-to-char-class.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+}
+
+function processExpressionCharset(expression, charset) {
```
DmitrySoshnikov: Let's call it as a predicate, with `should` prefix: `shouldProcessExpressionCharset` (otherwise `processExpressionCharset` sounds like action method).

src/optimizer/transforms/group-single-chars-to-char-class.js
```diff
@@ -0,0 +1,54 @@
...
+  const {type} = expression;
+
+  if (type === 'Disjunction') {
+    const {left, right} = expression;
+
+    return processExpressionCharset(left, charset)
+      && processExpressionCharset(right, charset);
+  }
+  else if (type === 'Char') {
```
DmitrySoshnikov: nit: on previous line if no comment:

```js
} else if (type === 'Char') {
  ...
}
```

With a comment might be:

```js
}

// Some else block.
else if (type === 'Char') {
  ...
}
```

silentroach: you should put eslint rules to repository :)
I will fix it today

DmitrySoshnikov: Yeah, setting up ESLint would be great (and I'll greatly appreciate a PR for this of course :))

silentroach:
all codestyle comments fixed

src/optimizer/transforms/group-single-chars-to-char-class.js
```diff
@@ -0,0 +1,56 @@
...
+    return shouldProcessExpressionCharset(left, charset)
+      && shouldProcessExpressionCharset(right, charset);
+  } else if (type === 'Char') {
+    const {value} = expression;
+
+    charset.set(value, expression);
+
+    return true;
+  }
+}
```
DmitrySoshnikov: Should we `return false;` at the end? Since other types are not considered, it'll return `undefined` (which technically works, but will be bad once we'll introduce type safety).

src/optimizer/transforms/group-single-chars-to-char-class.js
```diff
@@ -0,0 +1,56 @@
...
+
+    const characterClass = {
+      type: 'CharacterClass',
+      expressions: Array.from(charset.values())
+    };
+
+    if (node.capturing) {
+      node.expression = characterClass;
+    }
+    else {
```
DmitrySoshnikov: nit: on previous line

```js
} else {
```

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+
+[*.js]
+indent_style = space
+indent_size = 2
```
DmitrySoshnikov: Thanks, appreciated!

DmitrySoshnikov:
Great job, thanks again!

PR #72, opened 2017-05-05
better disjunction processor

silentroach:
Sorry, just forgot about that simple case: `/a|b|c/` -> `/[abc]/`.

Also here is another feature - group merge: `/[12]|3|4/` -> `/[1234]/`

And also chars are now sorted, so `/[43]|a|b|[53]|c|[9]/` will be optimized to `/[3459abc]/`

### DIFF
--- a/src/optimizer/transforms/__tests__/group-single-chars-to-char-class-test.js
+++ b/src/optimizer/transforms/__tests__/group-single-chars-to-char-class-test.js
@@ -10,6 +10,20 @@ const singleCharsGroupToCharClass = require('../group-single-chars-to-char-class
 
 describe('(a|b|c) -> ([abc])', () => {
 
+  it('replaces single chars disjunction to char class', () => {
+    const re = transform(/a|b|c/, [
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe('/[abc]/');
+  });
+
+  it('merges single chars disjunction with character class', () => {
+    const re = transform(/[43]|a|b|[53]|c|[9]/, [
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe('/[3459abc]/');
+  });
+
   it('replaces single chars group to char class', () => {
     const re = transform(/(a|b|c)/, [
       singleCharsGroupToCharClass,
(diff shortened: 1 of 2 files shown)


src/optimizer/transforms/__tests__/group-single-chars-to-char-class-test.js
```diff
@@ -10,6 +10,20 @@ const singleCharsGroupToCharClass = require('../group-single-chars-to-char-class
...
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe('/[abc]/');
+  });
+
+  it('merges single chars disjunction with character class', () => {
+    const re = transform(/[43]|a|b|[53]|c|[9]/, [
+      singleCharsGroupToCharClass,
+    ]);
+    expect(re.toString()).toBe('/[3459abc]/');
```
DmitrySoshnikov: Very nice!

src/optimizer/transforms/group-single-chars-to-char-class.js
```diff
@@ -8,49 +8,74 @@
...
+        .sort()
+        .map(key => charset.get(key))
     };
 
+    handlers[parent.type](path.getParent(), characterClass);
+  }
+};
+
+const handlers = {
+  RegExp: (path, characterClass) => {
```
DmitrySoshnikov: Let's use method notation on objects (to preserve `this` access; otherwise `this` is global in this case since captured lexically by arrow function)

src/optimizer/transforms/group-single-chars-to-char-class.js
```diff
@@ -8,49 +8,74 @@
...
+    const {node} = path;
+
+    node.body = characterClass;
+  },
+
+  Group: (path, characterClass) => {
```
DmitrySoshnikov: ditto, method

```js
Group(path, characterClass) {
  ...
},

File: src/optimizer/transforms/group-single-chars-to-char-class.js [hunk 4]
```diff
@@ -8,49 +8,74 @@
...
+    const {node} = path;
+
     if (node.capturing) {
       node.expression = characterClass;
     } else {
       path.replace(characterClass);
     }
   }
-}
+};
```
DmitrySoshnikov: nit: trailing comma

DmitrySoshnikov:
@silentroach, would you like to proceed with this PR? Or I can merge it, and do a follow up with fixes.

silentroach:
Yep, I'll fix it today, sorry for delay

silentroach:
fixed

DmitrySoshnikov:
Awesome stuff, thanks!

PR #73, opened 2017-05-06
check for correct boundaries in char class range

silentroach:
refs #1

### DIFF
--- a/src/parser/regexp.bnf
+++ b/src/parser/regexp.bnf
@@ -175,6 +175,15 @@ function getRange(text) {
 }
 
 /**
+ * Checks class range
+ */
+function checkClassRange(from, to) {
+  if (from > to) {
+    throw new SyntaxError(`Range ${from}-${to} out of order in character class`);
+  }
+}
+
+/**
  * Creates a character node.
  */
 function Char(value, kind, loc) {
@@ -694,6 +703,8 @@ NonemptyClassRanges
 
   | ClassAtom DASH ClassAtom ClassRanges
     {
+      checkClassRange($1.value, $3.value);
+
       $$ = [
         Node({
           type: 'ClassRange',
@@ -715,6 +726,8 @@ NonemptyClassRangesNoDash
 
   | ClassAtomNoDash DASH ClassAtom ClassRanges
     {
+      checkClassRange($1.value, $3.value);
+
       $$ = [
         Node({
           type: 'ClassRange',
--- a/src/parser/__tests__/parser-test262-test.js
+++ b/src/parser/__tests__/parser-test262-test.js
@@ -33,7 +33,8 @@ describe('test262', () => {
 
   it('invalid char range', () => {
     // 15.10.2.15-6-1
-    // TODO: /^[z-a]$/ - should throw, 'a' is less than 'z'
+    invalid('/^[z-a]$/', 'out of order');
+    invalid('/^[-z-a]$/', 'out of order');
   });
 
   it('invalid quantifier range', () => {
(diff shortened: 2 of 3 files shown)


hg42:
I think the parser should be tolerant or have a tolerant mode.
Otherwise we cannot write transformations that correct such things.

Perhaps such checks should be moved to a separate function like `check(...)`. Or `optimize` could have such rule sets.

silentroach:
why? you just can't use `[z-a]` in your code, it will not compile

hg42:
the parser can be used to write regexp extensions. An extension could allow this (and swap those characters).

But my concern is not about this particular case.
I just wanted to express, that a parser for general purposes should not be too picky, because it can be used for many things that may eventually break some rules.

I would put such a check in it's own pass or it could be placed in a generator for a specific target, because such rules depend on the target.

Example: you can parse named groups, even if the target does not allow them as long as you transform them to something else. If you would restrict the parser, you would not be able to do that.

You can also allow several syntax variants for named groups:
`(?<name>...)`, `(?P<name>...)`, `(?'name'...)` and generate whatever is supported by the target.

hg42:
also, if the check is placed in the parser, it only checks the input. But you may create such illegal constructs in transformations. So it would be better to check the output. Seems like the generator would be the best place. A separate checking pass could be even better, because you could react on such failures very differently. May be you want to explicitely warn about it but correct the error.

Throwing an exception also stops processing.
In some cases you might want to report all errors instead of only the first.
It's not fatal to the parsing process, so why stop it?

DmitrySoshnikov:
@hg42, yeah, I agreed on specifically extensions, that parser should not break them, however, in this specific case it's per ES-spec (it should be a parse error), and the `[z-a]` doesn't make sense, same as `{5,1}` for quantifier.

> Throwing an exception also stops processing.
In some cases you might want to report all errors instead of only the first.
It's not fatal to the parsing process, so why stop it?

I have a task opened in `Syntax` to implement error recovery: https://github.com/DmitrySoshnikov/syntax/issues/28, but for now it's not supported. Later we'll be able to collect all errors in one pass, and report them after parsing.

Anyways, for this specific case, we should definitely merge it, since it's per ES-spec.

src/parser/regexp.bnf
```diff
@@ -694,6 +703,8 @@ NonemptyClassRanges
 
   | ClassAtom DASH ClassAtom ClassRanges
     {
+      checkClassRange($1.value, $3.value);
```
DmitrySoshnikov: 👍 We should also call this check for the `NonemptyClassRangesNoDash : ClassAtomNoDash DASH ClassAtom ClassRanges` production below.

You can test this by two examples:

```
[a-z]
[-a-z]
```

hg42:
I think it should be possible to get a list of errors **and** the AST in such cases.
E.g. for analyzers, linters or for teaching purposes.
Only errors fatal for parsing (like structural errors) should stop parsing.
There are cases when non-programmers make errors that can be corrected, then I would want to ask them and display corresponding parts of the tree (I'm talking about real cases I had in my jobs).

src/parser/regexp.bnf
DmitrySoshnikov: @silentroach, would you like to proceed with this one? Or I can merge this request, and do a follow up with fixing this part.

silentroach: I'll fix it of course. Will try to do it today

silentroach: fixed

DmitrySoshnikov:
Perfect; thanks! 👍

PR #77, opened 2017-06-28
Correctly escape front and backslash

### LINKED ISSUE #76: Optimizer: Do not optimize away escaped back and forward slashes
Currently, the regex `/^.\//` is being optimized to `/^.//` which is not a valid regex.

BrainMaestro:
Fixes #76


DmitrySoshnikov:
Good catch; thanks!

Issue #78, opened 2017-06-29
Separate \w, \d, ... into separate node type


RReverser:
All the other character types allow to match exactly a single character (as the node type `Character` suggests), but `\w`, `\d`, ... are mixed with single-char escapes under `meta` kind.

It seems reasonable to separate them into their own node type as they're something very different than a single char.

DmitrySoshnikov:
Yes, potentially we can move them to a separate node type. Since it'll be a breaking change, we'll need to have a good use-case for it. Initially I tried to make a simple AST format, but I'm open to reconsider it if there is a use-case (we'll also need to update all the docs, plugins, etc).

RReverser:
Mostly useful for analysis of a regexp (complexity and such). It's useful to rely on `Character` to be a single character, no matter what kind it is, while these classes break this assumption.

And just in general, it seems semantically correct to have them as a different node type.

DmitrySoshnikov:
OK, I think it might make sense, I think we can go with this (that's said, we'll need to update all the related docs, etc).

PR #79, opened 2017-06-29
Add initial TypeScript definitions

RReverser:
When published to npm, this will help TS developers to automatically get strong typings when importing a module.

Covering only parser for now as that's the most interesting for me, but I'm sure someone else can pick it up and add other APIs.

Semi-related to #34.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -36,19 +36,19 @@ declare namespace Node {
   }
 
   interface Disjunction extends Base<'Disjunction'> {
-    expressions: Expression[];
+    expressions: (Expression | null)[];
   }
 
   interface CapturingGroup extends Base<'Group'> {
     capturing: true;
     number: number;
     name?: string;
-    expression: Expression;
+    expression: Expression | null;
   }
 
   interface NoncapturingGroup extends Base<'Group'> {
     capturing: false;
-    expression: Expression;
+    expression: Expression | null;
   }
 
   type Group = CapturingGroup | NoncapturingGroup;
@@ -93,7 +93,7 @@ declare namespace Node {
   interface LookaroundAssertion extends Base<'Assertion'> {
     kind: 'Lookahead' | 'Lookbehind';
     negative?: true;
-    assertion: Expression;
+    assertion: Expression | null;
   }
 
   type Assertion = SimpleAssertion | LookaroundAssertion;
@@ -109,7 +109,7 @@ declare namespace Node {
     | Assertion;
 
   interface RegExp extends Base<'RegExp'> {
-    body: Expression;
+    body: Expression | null;
     flags: string;
   }
 }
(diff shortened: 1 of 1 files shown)


RReverser:
Just realized that many nodes can in fact be `null`s apparently...

DmitrySoshnikov:
@RReverser, this looks perfect, I was thinking actually to add TypeScript's annotations to the whole code.

> Just realized that many nodes can in fact be nulls apparently...

Do we need to do something in addition here?

RReverser:
> Do we need to do something in addition here?

Yeah, unless you want to make #80 change, I'd add explicit `| null` to all the places where it's possible - TypeScript can distinguish this in a `strict` mode, making typechecks stronger and ensuring that a developer handles `null` case before dereferencing as `node.type` or something.

DmitrySoshnikov:
Yeah, I'm not sure about #80 yet. The `null` is actively used in estree spec, and e.g. Babylon parsers `var a;` with `null` for `init`. The explicit `{type: 'Empty'}` also needs a good use-case. So for now I think `null` makes sense.

RReverser:
@DmitrySoshnikov Oh ok, it was unclear to me that regexp-tree intends to match the fashion of ESTree. It just seemed better to me because in regexp land "nothing" (`∅`) and "empty" (`ε`) languages are different things, and having a clarification in `type` could perhaps help with understanding whereas `null` is somewhat ambigous.

Although I guess regexp-tree parsers/transformers don't really return "match nothing" nodes?

DmitrySoshnikov:
@RReverser, good point on epsilon (`ε`), meaning "empty" string, i.e. not consuming input chars. For now I thin we can leave it as `null`, and extend the type declaration with `| null` where is needed. Later we can reconsider the "empty" type if there is a need.

(FYI, I'll be out on vacation, but will try to take a look at PRs/issues at some point).

RReverser:
@DmitrySoshnikov Ok, fair enough, I'll add `| null` later today.

RReverser:
Updated. Also, I noticed there is an undocumented `symbol` property on `Char` nodes which seems to be always `undefined` - is that intentional?

DmitrySoshnikov:
> Also, I noticed there is an undocumented symbol property on Char nodes which seems to be always undefined - is that intentional?

Yes, we should probably add `symbol` on simple chars as well for consistency (although, it'll be the same as `value`). The `symbol` was added for encoded number-chars.

DmitrySoshnikov:
Thanks for the PR; it's a good direction!

RReverser:
@DmitrySoshnikov 

> although, it'll be the same as value

Not for escaped chars like `\n` (another reason to make split as in #78 - those can't have `symbol` because they produce sets).

PR #82, opened 2017-07-10
Preserve escape for dot and closing brace

BrainMaestro:
Added some missing characters that were being optimized incorrectly (escape removed).


DmitrySoshnikov:
Good catch; Thanks!

Issue #92, opened 2017-11-05
Error thrown when using range \xA0-\uFFFF


Golmote:
When using range `[\xA0-\uFFFF]` in a character class, the parser throws the following error:

```
SyntaxError: Range \xA0-\uFFFF out of order in character class
```

I guess this type of range is unusual, but I'm pretty sure it's valid (and AFAICT it works in browsers).

I think the same error is thrown when combining a simple char with a char code, like `a-\xA0` or `a-\uFFFF`.

DmitrySoshnikov:
Yeah, we should check which values come here: https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/parser/regexp.bnf#L180

PR #96, opened 2017-11-06
Handle codePoints in checkClassRange()

Golmote:
This is a fix for #92.

I added a `codePoint` key on `Char` tokens, so that we can compare all syntaxes the same way.

I had to handle a special case for `control` chars, which seem to not be allowed in character ranges.

I also handled `meta` character classes short forms (`\w`, `\s`, etc.) using `codePoint` with the value `NaN`. I'm not really happy with this. In fact, something like `[a-\s]` should not even be parsed as a character class in the first place, but rather as three separate chars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,12 +100,16 @@ Which produces an AST node corresponding to this regular expression:
     left: {
       type: 'Char',
       value: 'a',
-      kind: 'simple'
+      symbol: 'a',
+      kind: 'simple',
+      codePoint: 97
     },
     right: {
       type: 'Char',
       value: 'b',
-      kind: 'simple'
+      symbol: 'b',
+      kind: 'simple',
+      codePoint: 98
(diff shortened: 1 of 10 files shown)


Golmote:
Ogod what have I done! Is it ok to add this property on Chars? I'll have to fix all tests... but I'll wait for your opinion first. ^^

DmitrySoshnikov:
This looks great, thanks! We should update the documentation in the main [README.md](https://github.com/DmitrySoshnikov/regexp-tree#char) (specification for the `Char` node, adding this `codePoint` property, and also mention it in all examples). We should also probably add `symbol` property to simple chars as well for consistency.

Golmote:
Ok, README updated!

DmitrySoshnikov:
OK, perfect!

PR #97, opened 2017-11-08
Remove duplicates in disjunctions Fix #94

### LINKED ISSUE #94: Optimizer: implement "remove duplicates" transform for disjunction
It would be nice if the optimizer removed duplicates in disjunctions:

`/ab|ab/ -> /ab/`

Golmote:
Hi! This is an attempt to fix #94.

The issue I'm facing is that it only replaces one duplicate at a time.
Thus calling `regexp-tree -e "/(ab|ab|ab|ab)/" -o` returns `/(ab|ab|ab)/`, which is kind of annoying.

Any idea to make it more useful?

### DIFF
--- a/src/optimizer/transforms/__tests__/disjunction-remove-duplicates-transform-test.js
+++ b/src/optimizer/transforms/__tests__/disjunction-remove-duplicates-transform-test.js
@@ -1,0 +1,53 @@
+/**
+ * The MIT License (MIT)
+ * Copyright (c) 2017-present Dmitry Soshnikov <dmitry.soshnikov@gmail.com>
+ */
+
+'use strict';
+
+const {transform} = require('../../../transform');
+const disjunctionRemoveDuplicates = require('../disjunction-remove-duplicates-transform');
+
+describe('(ab|bc|ab) -> (ab|bc)', () => {
+
+  it('removes duplicates from disjunction', () => {
+    const re = transform(/ab|bc|ab/, [
+      disjunctionRemoveDuplicates
+    ]);
+    expect(re.toString()).toBe('/ab|bc/');
+  });
+
+  it('preserves order', () => {
+    const re = transform(/bc|bc|ab/, [
+      disjunctionRemoveDuplicates
+    ]);
+    expect(re.toString()).toBe('/bc|ab/');
+  });
+
+  it('replaces disjunction containing only one part', () => {
+    const re = transform(/ab|ab/, [
+      disjunctionRemoveDuplicates
+    ]);
+    expect(re.toString()).toBe('/ab/');
+  });
+
+  it('does not merge capturing groups', () => {
+    const re = transform(/(ab)|(ab)/, [
+      disjunctionRemoveDuplicates
+    ]);
+    expect(re.toString()).toBe('/(ab)|(ab)/');
+  });
+
+  it('handles recursion', () => {
+    const re = transform(/(ab|bc|ab)|(bc|cd|bc)/, [
+      disjunctionRemoveDuplicates
+    ]);
+    expect(re.toString()).toBe('/(ab|bc)|(bc|cd)/');
+
+    const re2 = transform(/(?:ab|bc|ab)|(?:ab|bc|ab)/, [
+      disjunctionRemoveDuplicates
+    ]);
+    expect(re2.toString()).toBe('/(?:ab|bc)/');
+  });
+
+});
--- a/src/optimizer/transforms/disjunction-remove-duplicates-transform.js
+++ b/src/optimizer/transforms/disjunction-remove-duplicates-transform.js
@@ -1,0 +1,43 @@
+/**
+ * The MIT License (MIT)
+ * Copyright (c) 2017-present Dmitry Soshnikov <dmitry.soshnikov@gmail.com>
+ */
+
+'use strict';
+
+const NodePath = require('../../traverse/node-path');
+
+const {
+  disjunctionToList,
+  listToDisjunction,
+} = require('../../transform/utils');
+
+/**
+ * Removes duplicates from a disjunction sequence:
+ *
+ * /(ab|bc|ab)+(xy|xy)+/ -> /(ab|bc)+(xy)+/
+ */
+module.exports = {
+  Disjunction(path) {
+    const {node} = path;
+
+    // Make unique nodes.
+    const uniqueNodesMap = {};
+
+    const parts = disjunctionToList(node).filter(part => {
+      const encoded = NodePath.getForNode(part).jsonEncode();
+
+      // Already recorded this part, filter out.
+      if (uniqueNodesMap.hasOwnProperty(encoded)) {
+        return false;
+      }
+
+      uniqueNodesMap[encoded] = part;
+      return true;
+    });
+
+    // Replace with the optimized disjunction.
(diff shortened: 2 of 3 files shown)


DmitrySoshnikov:
@Golmote, thanks, I'll take a look closely into this. We should actually flatten the values from the disjunctions, collect them, and then build a new disjunction, excluding the duplicates.

E.g. this:

```
/ab|xy|ab|xy/
```

should be (order should be maintained):

```
/ab|xy/
```

Transforms can have `init` hook where we can initialize the set, and then maintain it:

```js
module.exports = {
  _parts: null,

  init() {
    this._parts = new Set();
  },

  Disjunction(path, v) {
    const {node} = path;

    // Already handled.
    if (this._parts.has(node.left)) {
      return;
    }

    // Extract all parts from the disjunction: /xy|ab|xy/ -> Set([xy, ab])
    this._extractParts(node);

    // At this point it should contain flattened parts.
    // console.log(this._parts);

    let newDisjunctionNode = null;

    for (const part of this._parts) {
      // Create recursively a new disjuntion node, checking
      // hasEqualSource, and excluding the duplicates.
    }

    path.replace(newDisjunctionNode);
  },

  _extractParts(node) {
    if (node.left.type === 'Disjunction') {
      this._extractParts(node.left);
    } else {
      this._parts.add(node.left);
    }
    this._parts.add(node.right);
  }
};
```

DmitrySoshnikov:
We even extract these useful util functions to the to the `src/transform/utils.js`, and then use as:

```js
const transformUtils = require('../transform/utils');

// Flattened list of disjunction items
// /a|b|c|d/ -> [a, b, c, d]
const list = transformUtils.disjunctionToList(complexDisjuction);

// Builds recursively a Disjunction node from the list
// [a, b, c, d] -> {{{a, b}, c}, d}
const disjunction = transformUtils.disjunctionFromList(list);
```

These util functions might be reused further in other transforms.

DmitrySoshnikov:
I added transform utils in https://github.com/DmitrySoshnikov/regexp-tree/commit/adf103a9efff91eea8242ecb935f6c0406d02d52. So we can do:

```js
const filteredList = transformUtils.disjunctionToList(disjunction).filter(node => {
  // maintain set, and filter out by the same source
});

const newDisjunction = transformUtils.listToDisjunction(filteredList);

path.replace(newDisjunction);
```

I'll take a look later at transform, and add here in case.

Golmote:
I made a new version. But it's acting weird. When processing this regexp: `/ab|bc|ab/`, the disjunction is properly replaced with one with unique parts...
But it looks like the former `left` part gets visited next... but it should have been deleted by the `replace()` call, shouldn't it?

I feel like I'm missing something obvious, I don't know...

Golmote:
I just updated this PR to use your functions, but the issue I described above is still occurring.

Golmote:
With the current code of the PR, `/ab|bc|ab/` gives `/ab|bc|bc/`. It's like the left part (`ab|bc`) of the initial disjunction was visited even after the entire node has been replaced with `ab|bc`... replacing the latter's left part with `ab|bc` again. I can't get a grip on the behaviour.

DmitrySoshnikov:
Oh, I see what's the issue: we replace a disjunction, _with another disjunction_ -- and exactly the new disjunction is visited again.

Anyways, to solve this, we need to stop traversal, once we have handled this disjunction, and don't go further.

This should work (please maintain the docblocks, etc):

```js
/**
 * The MIT License (MIT)
 * Copyright (c) 2017-present Dmitry Soshnikov <dmitry.soshnikov@gmail.com>
 */

'use strict';

const NodePath = require('../../traverse/node-path');

const {
  disjunctionToList,
  listToDisjunction,
} = require('../../transform/utils.js');

/**
 * Removes duplicates from a disjunction sequence:
 *
 * /(ab|bc|ab)+(xy|xy)+/ -> /(ab|bc)+(xy)+/
 */
module.exports = {
  Disjunction(path) {
    const {node} = path;

    // Flatten the disjunction to list.
    let parts = disjunctionToList(node);

    // Make unique nodes.
    const uniqueNodesMap = {};

    parts = parts.filter(part => {
      const encoded = NodePath.getForNode(part).jsonEncode();

      // Already recorded this part, filter out.
      if (uniqueNodesMap.hasOwnProperty(encoded)) {
        return false;
      }

      uniqueNodesMap[encoded] = part;
      return true;
    });

    // Replace with the optimized disjunction.
    path.replace(listToDisjunction(parts));

    // Since we replace this disjunction with the new disjunction,
    // we have to stop traversal. Otherwise, it'll go recursively
    // to the new disjunction.
    return false;
  }
};
```

Once you'll have checked it works, please also add a unit test (with several use-cases) to the appropriate `__tests__` directory of the disjunction.

DmitrySoshnikov:
Note, I added ability to stop traversal in https://github.com/DmitrySoshnikov/regexp-tree/commit/bdd58628ede3eb972b2ed05e1a5d3404dec9a60a, so you'll have to rebase.

Golmote:
@DmitrySoshnikov I'm sorry to insist but I think you you're wrong: it's not the new disjunction that's being visited, it's the old one's `left` part.
I just tested with `/ab|ab|bc/`. First the top-level disjunction is processed and replaced with a new `ab|bc` disjunction. But then, a disjunction `ab|ab` is visited somehow (and I think it has the new disjunction as its parent).

Furthermore, we can't just stop the traversal. We do want to have recursion for cases like
`/(ab|bc|ab)|(bc|cd|bc)/` -> `/(ab|bc)|(bc|cd)/`
or even
`/(?:ab|bc|ab)|(?:ab|bc|ab)/` -> `/(?:ab|bc)/`

DmitrySoshnikov:
@Golmote, interesting; OK, I'll double-check, perhaps the traversal already started traversing the sub-object, and somehow got a reference to it before we replaced the top-level disjunction. I'll get back to it later, at the meantime if you have ideas, let me know.

Ideally, in this specific case it'd be easier to traverse bottom-up, i.e. child disjunctions first, and then going up, replacing top levels.

Alternatively, we could handle top-level `RegExp` node, and manually search for `Disjunction`s in it.

DmitrySoshnikov:
OK, that is it -- the old node was "cached", and the traversal continued for it. So since a node can be replaced during the traversal, we have to recalculate the node on further traversal. Also, you are right, we don't need to stop the traversal to handle nested cases.

I submitted a fix for the traversal in https://github.com/DmitrySoshnikov/regexp-tree/commit/36d6dde8eab493cb10afead7fccedfdd7faa8a9f.

Please check with the same plugin code from above (just remove the `return false` since we don't need to stop the traversal).

DmitrySoshnikov:
Actually, some tests are failing after 36d6dde. Looking into it...

DmitrySoshnikov:
OK, tests are fixed in https://github.com/DmitrySoshnikov/regexp-tree/commit/43975412a3c8fa7e4bef65aaf69546dbdf82fa1a. So we should be good now :)

Golmote:
Yay! Thank you very much for digging into this! The PR is updated with the last version of the transform and a test files with several use-cases.

DmitrySoshnikov:
@Golmote, awesome! Thanks; and also for the all the generic tool bugs caught in this PR.

PR #99, opened 2017-11-12
Add support for quantifiers merge in Optimizer

Golmote:
This transform will merge siblings quantifiers.

`/a+a+a+/` -> `/a{3,}/`
`/a+a*/` ->` /a+/`
`/a{4,5}a{2}a{1}/` -> `/a{7,8}/`
`/a??a??/` -> `/a{0,2}?/`

The process is aborted when the two quantifiers have a different greediness and none of them is an open greedy quantifier, because I couldn't think of a reliable way to merge them (if there is...):
`/a{1,2}?a{1,2}/` is left untouched (since it has not the same behaviour as `/a{2,4}/` nor `/a{2,4}?/`)


DmitrySoshnikov:
Whoa! Great job on all these PRs. I'll take a detailed look later on today.

Golmote:
Thanks! I know there a lot of tests in there, but this should be an exhaustive list of the cases.

Also, I'm realising right now that this transform will make the [char-star-to-plus transform](https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/optimizer/transforms/char-star-to-plus-transform.js) obsolete.

DmitrySoshnikov:
Yep, I'm gonna look at the tests and code, and yes, we can remove the `char-star-to-plus transform` after this merge.

DmitrySoshnikov:
Actually, feel free to remove `char-star-to-plus transform` in this PR as well.

DmitrySoshnikov:
OK, the code looks good overall; fantastic plugin, thanks!

DmitrySoshnikov:
BTW, we can probably also use range quantifiers to transform this `/aaaaa/` to `/a{5}/`.

Golmote:
@DmitrySoshnikov Agreed, that would be great! I'll open an issue to discuss it first, though.

PR #100, opened 2017-11-12
Optimizer: don't unescape \^ in character class

Golmote:
We don't want `/[\^a]/` to be optimized to `/[^a]/`.


DmitrySoshnikov:
Good catch!

PR #101, opened 2017-11-12
Optimizer should run until it finds the best optimization

Golmote:
Currently, optimizing `/a|a/` gives `/[a]/` and optimizing again `/[a]/` gives `/a/`.
We should get the most optimized result directly.

This PR lets the optimizer run until it finds the best optimization. The drawback is that it has to run minimum twice to know...

### DIFF
--- a/src/optimizer/__tests__/optimizer-integration-test.js
+++ b/src/optimizer/__tests__/optimizer-integration-test.js
@@ -73,4 +73,12 @@ describe('optimizer-integration-test', () => {
       .toBe(optimized.toString());
   });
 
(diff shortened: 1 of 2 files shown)


DmitrySoshnikov:
For this specific case we could solve it by just reordering the transforms (they are applied in pipeline in order). So we could put the https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/optimizer/transforms/index.js#L19 at the very end.

But I actually like the approach, it'll save us from the reordering issues in the future, and it's fine to run in a loop in this case. Great job!

PR #102, opened 2017-11-12
Optimizer: remove unnecessary groups

Golmote:
This transform removes unnecessary groups:

`/a(?:bc)d/` -> `/abcd/`
`/(?:ab|bc)/` -> `/ab|bc/`
`/(?:a)+/` -> `/a+/`

It does not remove capturing groups, nor disjunctions other than top-level ones, nor complex groups with quantifier:
`/a(bc)d/`, `/a(?:ab|bc)d/`, `/(?:ab)+/` are left unchanged.

### DIFF
--- a/src/optimizer/transforms/ungroup-transform.js
+++ b/src/optimizer/transforms/ungroup-transform.js
@@ -1,0 +1,41 @@
+/**
+ * The MIT License (MIT)
+ * Copyright (c) 2017-present Dmitry Soshnikov <dmitry.soshnikov@gmail.com>
+ */
+
+'use strict';
+
+/**
+ * A regexp-tree plugin to remove unnecessary groups.
+ *
+ * /(?:a)/ -> /a/
+ */
+
+module.exports = {
+  Group(path) {
+    const {node, parent} = path;
+    const childPath = path.getChild();
+
+    if (node.capturing) {
+      return;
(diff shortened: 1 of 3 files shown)


src/optimizer/transforms/ungroup-transform.js
```diff
@@ -0,0 +1,40 @@
...
+    const {node} = path;
+
+    if (node.capturing) {
+      return;
+    }
+
+    if (
+      path.getChild() &&
+      path.getChild().node.type === 'Disjunction' &&
+      path.getParent().node.type !== 'RegExp'
```
DmitrySoshnikov: Can simplify using just `parent` reference from `path`.

```js
const {node, parent} = path;
...
parent.type !== 'RegExp'
```

src/optimizer/transforms/ungroup-transform.js
```diff
@@ -0,0 +1,40 @@
...
+    ) {
+      return;
+    }
+
+    if (
+      path.getParent().node.type === 'Repetition' &&
```
DmitrySoshnikov: ditto, `parent`

Golmote:
Changes done!
I also added a variable holding the result of `getChild()` since there is no need to rebuild it it so many times.

PR #103, opened 2017-11-14
Optimizer: don't throw errors on empty groups

Golmote:
This PR prevents the optimizer from throwing errors when handling empty groups or empty parts:

Optimizing `/a(?:)b/` used to throw in the `ungroup` transform. Now it leaves the regexp as it is.
Optimizing `/a||/` used to throw in the `disjunction-remove-duplicates` transform. Now it handles it properly and optimizes as `/a|/`.
Optimizing `/a|b|/` used to throw in the `group-single-chars-to-char-class` transform. Now it leaves the regexp as it is.

In the last case, I wonder if `[ab]?` would be an appropriate optimization. What is your opinion on this @DmitrySoshnikov ?

Also, I intend to add a specific transform to handle empty (non-capturing) groups at some point.

### DIFF
--- a/src/optimizer/transforms/__tests__/disjunction-remove-duplicates-transform-test.js
+++ b/src/optimizer/transforms/__tests__/disjunction-remove-duplicates-transform-test.js
@@ -50,4 +50,11 @@ describe('(ab|bc|ab) -> (ab|bc)', () => {
     expect(re2.toString()).toBe('/(?:ab|bc)/');
   });
 
+  it('handles empty parts', () => {
+    const re = transform(/a|b|||/, [
+      disjunctionRemoveDuplicates
+    ]);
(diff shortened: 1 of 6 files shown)


DmitrySoshnikov:
Sorry for the delayed reply, and thanks for the fix!

As for `/a|b|/` ->  `[ab]?` -- not sure yet, potentially yes: we should double-check whether it changes the semantics.

Golmote:
Quick followup on the `/a|b|/` -> `[ab]?` question: AFAICT it does not change the semantics when the empty part is in the last position. I does change it in every other cases though: `/a||b/` -/> `[ab]?`

I don't think it's worth making an optimization for this.

DmitrySoshnikov:
@Golmote , OK, sounds good, thanks for the clarifications.

PR #105, opened 2017-11-17
Traverse: allow explicit stop of the traversal when using catch-all handler

Golmote:
This is just to be consistent with per-node handlers.


DmitrySoshnikov:
Good catch!

PR #106, opened 2017-11-17
Traverse: Delete node from NodePath registry when removed

Golmote:
This fixes a weird bug I had with nodes existing in Registry being `null`.


Golmote:
I think it can be reproduced when referencing a node formerly used in a path that was deleted. Something like:

```js
const {node} = path;
path.remove();
otherPath.replace(node);
```

DmitrySoshnikov:
Yep, also a good catch, thanks!

PR #107, opened 2017-11-17
Combine repeating patterns transform

Golmote:
This transform combines repeating patterns in alternatives. It is divided in three parts:

Look for two repeating patterns and combine them in a Repetition:
`/abcabc/` -> `/(?:abc){2}/`

Look for a pattern existing in a Repetition on the left:
`/(?:abc){2}abc/` -> `/(?:abc){3}/`

Look for a pattern on the left that matches a Repetition:
`/abc(?:abc){2}/` -> `/(?:abc){3}/`

**Note that this PR depends on (and includes) PR #106.**

### DIFF
--- a/src/optimizer/transforms/combine-repeating-patterns-transform.js
+++ b/src/optimizer/transforms/combine-repeating-patterns-transform.js
@@ -1,0 +1,206 @@
+/**
+ * The MIT License (MIT)
+ * Copyright (c) 2017-present Dmitry Soshnikov <dmitry.soshnikov@gmail.com>
+ */
+
+'use strict';
+
+const NodePath = require('../../traverse/node-path');
+
+/**
+ * A regexp-tree plugin to combine repeating patterns.
+ *
+ * /^abcabcabc/ -> /^abc{3}/
+ * /^(?:abc){2}abc/ -> /^(?:abc){3}/
+ * /^abc(?:abc){2}/ -> /^(?:abc){3}/
+ */
+
+module.exports = {
+  Alternative(path) {
+    const {node} = path;
+
+    // We can skip the first child
+    let index = 1;
+    while (index < node.expressions.length) {
+      let child = path.getChild(index);
+      index = Math.max(1, combineRepeatingPatternLeft(path, child, index));
+
+      if (index >= node.expressions.length) {
+        break;
+      }
+
+      child = path.getChild(index);
+      index = Math.max(1, combineWithPreviousRepetition(path, child, index));
+
+      if (index >= node.expressions.length) {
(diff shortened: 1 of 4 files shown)


src/optimizer/transforms/combine-repeating-patterns-transform.js
```diff
@@ -0,0 +1,206 @@
...
+      });
+    } else {
+      rightLength = 1;
+      left = alternative.getChild(index - 1);
+      if (left.node.type === 'Group' && !left.node.capturing) {
+        left = left.getChild();
+      }
+    }
+
+    if (left.hasEqualSource(right)) {
```
DmitrySoshnikov: BTW, later we should probably start caching the sources of the NodePaths in some registry for performance improvements (and checking the cached source in `hasEqualSource` implementation).

src/optimizer/transforms/combine-repeating-patterns-transform.js
```diff
@@ -0,0 +1,206 @@
...
+
+  return index;
+}
+
+// (?:abc){2}abc -> (?:abc){3}
+function combineWithPreviousRepetition(alternative, child, index) {
+  const {node} = alternative;
+
+  let i = 0;
+  while (i < index) {
```
DmitrySoshnikov: Amount of loops is growing in the optimizer module :) Should be fine for now, but later we should be checking performance regressions. Usually it's OK on a smaller code base, but when you have thousand of files which you need to handle, smaller peaces can combine to some performance bottleneck.

DmitrySoshnikov:
Published in `0.0.66`.

src/optimizer/transforms/combine-repeating-patterns-transform.js
Golmote: @DmitrySoshnikov Aren't you afraid it will be hard to keep the source in sync with the node? I mean, changing any property of the node will change the source.

src/optimizer/transforms/combine-repeating-patterns-transform.js
Golmote: @DmitrySoshnikov: I agree. It's super hard to make smart optimizations while keeping the amount of loops low :D

src/optimizer/transforms/combine-repeating-patterns-transform.js
DmitrySoshnikov: @Golmote, yeah, good point, this can quickly go unsync. For now that's said it should be fine in general, and we can just cache the source into variables in case in transforms themselves (and recalculate it if nodes change).

PR #108, opened 2017-11-18
Optimizer: Escape $ . / { } | in char-class-to-single-char transform

Golmote:
The `char-class-to-single-char` transform must escape `$`, `.`, `/`, `{`, `}` and `|` when turning them into single chars.

`/[$]/`-> `/\$/`

Also, I wondered about the `$` char in thoses regexps: [L61](https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/optimizer/transforms/char-class-to-single-char-transform.js#L61) and [L65](https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/optimizer/transforms/char-class-to-single-char-transform.js#L65). What is it for?

### DIFF
--- a/src/optimizer/transforms/char-class-to-single-char-transform.js
+++ b/src/optimizer/transforms/char-class-to-single-char-transform.js
@@ -67,6 +67,8 @@ function getInverseMeta(value) {
     : value.toLowerCase();
 }
 
+// Note: \{ and \} are always preserved to avoid `a[{]2[}]` turning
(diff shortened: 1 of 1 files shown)


DmitrySoshnikov:
Hm, I guess the code for meta was just reused for `$`, can't recall what was the motivation. Thanks for the fix though! If you figure out the `$`, feel free to change it as you think might fit better.

PR #109, opened 2017-11-18
Transform: Prevent disjunctionToList from throwing on /|/

Golmote:
This PR prevents `disjunctionToList` from throwing an error when processing `/|/`.


DmitrySoshnikov:
Woops! Good catch, thanks!

Issue #81, opened 2017-06-30
\s optimization changes runtime semantics


RReverser:
As per ECMAScript spec, `\s` matches any WhiteSpace or LineTerminator character:

![1248bc49-5e3a-45b7-aa45-8426aa28fd0a](https://user-images.githubusercontent.com/557590/27735672-bfa8e6b8-5d98-11e7-8352-cadcaa7ab926.png)

Currently optimizer docs say it transforms `[ \t\r\n\f]` to `\s` (and the opposite), but this changes runtime semantics of the RegExp as in fact `\s` matches many more characters:

![e42aae83-29e9-47ad-982e-e49315c9db56](https://user-images.githubusercontent.com/557590/27735676-c0c31168-5d98-11e7-9eac-a7b6c0aace96.png)
![94cf74b5-69da-47e2-a7d5-751b30af462d](https://user-images.githubusercontent.com/557590/27735678-c4663b88-5d98-11e7-9dc0-c866003c3fda.png)

I don't know if this is an intentional deviation and these are considered as edge cases that should be ignored, but it would be nice to make such "loose" transformations at least optional under a flag, as otherwise an optimized regexp can behave differently than original one, which is dangerous to use as an ESLint plugin or a transpiler.

DmitrySoshnikov:
Good catch! In practice though, as you mentioned, I'd consider it as an edge case. The "loose" mode sounds interesting, or we can just handle a white/black list of needed transforms, and exclude some transform if it changes the semantics on practice.

RReverser:
I just wonder if people ever put such an explicit list of whitespace characters but mean `\s`, or maybe safer to just remove this particular optimisation and assume people wrote this list of characters because they intended to? (up to you to decide, of course)

mathiasbynens:
Yeah, let’s not change runtime semantics.

DmitrySoshnikov:
OK, need to think about it. Depending on how often the use-case is, we can either exclude the `\s` transform by default (and allow to opt-in), or, if it's very rare, to keep, and allow to opt-out when needed.

RReverser:
> or, if it's very rare, to keep, and allow to opt-out when needed

I don't think keeping it as-is is a very good idea, *especially* if it's rarely used, since then it can just bite silently much later in runtime.

PR #112, opened 2017-11-19
Optimizer: Add whitelist parameter to optimize()

Golmote:
This PR adds a `whitelist` parameter to the `optimize` method, as suggested in #81.

This could be a first step towards a list of optional ("loose") transforms that can only be performed when explicitely added to the whitelist.

I based this exactly on the compat-transpiler behaviour.
Alternatively, since it might be tedious to list all optimization transforms, we might want to consider using an object instead, with transform names as keys and booleans as values.

### DIFF
--- a/index.js
+++ b/index.js
@@ -120,8 +120,8 @@ const regexpTree = {
    *
    * @return TransformResult object
    */
-  optimize(regexp) {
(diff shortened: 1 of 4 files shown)


DmitrySoshnikov:
OK, sounds good; thanks!

PR #113, opened 2017-11-19
Optimizer: Remove $ from list of metas in char-class-to-single-car transform

Golmote:
Following the discussion in #108, I'm pretty sure this `$` character is a typo. I can't think of any use case where we'd want to consider `$` as a meta character that can be inversed.


DmitrySoshnikov:
OK, hopefully it was just my typo 😄 (just appeared in two places by some reason, which I cannot recall). Thanks!

PR #115, opened 2017-11-19
Optimizer: convert char class to \w differently depending on i and u flags

Golmote:
The meaning of `\w` is extended when regexp has `i` and `u` flags so we should not do the optimization in this case.

### DIFF
--- a/src/optimizer/transforms/char-class-to-meta-transform.js
+++ b/src/optimizer/transforms/char-class-to-meta-transform.js
@@ -10,13 +10,19 @@
  * their meta symbols equivalents.
  */
 module.exports = {
+  _hasIFlag: false,
+  _hasUFlag: false,
+  init(ast) {
+    this._hasIFlag = ast.flags.includes('i');
+    this._hasUFlag = ast.flags.includes('u');
+  },
   CharacterClass(path) {
 
     // [0-9] -> \d
     rewriteNumberRanges(path);
 
     // [a-zA-Z_0-9] -> \w
-    rewriteWordRanges(path);
+    rewriteWordRanges(path, this._hasIFlag, this._hasUFlag);
 
     // [ \t\r\n\f] -> \s
     rewriteWhitespaceRanges(path);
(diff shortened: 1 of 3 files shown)


mathiasbynens:
With `i` and `u` flags enabled, you could do the same optimization for `[0-9a-zA-Z_\u017F\u212A]` specifically.

Golmote:
@mathiasbynens Really? Wow I didn't think there were only two exceptions!
I've updated the PR to handle those.

I've also added the case where, the `i` flag being set, there is no need for both `a-z` and `A-Z`.

So:
`/[\da-zA-Z_$]/`-> `/[\w$]/`
`/[\da-z_$]/i`-> `/[\w$]/i`
`/[\dA-Z_$]/i`-> `/[\w$]/i`
`/[\da-z_\u017F\u212A$]/iu`-> `/[\w$]/iu`
`/[\dA-Z_\u{017F}\u{212A}$]/iu`-> `/[\w$]/iu`

mathiasbynens:
More could potentially be added as the Unicode Standard gets updates, but yeah, for now it’s only those two. https://mathiasbynens.be/notes/es6-unicode-regex#impact-i has some more info.

src/optimizer/transforms/char-class-to-meta-transform.js
```diff
@@ -10,13 +10,21 @@
  * their meta symbols equivalents.
  */
 module.exports = {
+  _hasIFlag: false,
+  _hasUFlag: false,
+  shouldRun(ast) {
```
DmitrySoshnikov: In this case you uses it to do initialization rather than the actual check whether the plugin should run. In this case there is `init` hook for this:

```js
init() {
  this._hasIFlag = ast.flags.includes('i');
  this._hasUFlag = ast.flags.includes('u');
}
```

Golmote: You're right! I got lazy when I realized the `init` function was called without arguments. Fixed this!

DmitrySoshnikov: OK, perfect!

PR #116, opened 2017-11-19
Parser: Parse surrogate pairs and es6 unicode code point escapes only when `u` flag is set

Golmote:
I had to mess up with the BNF grammar more than I expected, but it appears to work as intended.


src/parser/regexp.bnf
```diff
@@ -35,96 +37,112 @@ GROUP_NAME              ([\w$]|\\'u'[0-9a-fA-F]{4}|\\'u{'[0-9a-fA-F]{1,}'}')+
...
+'.'                                 return 'ANY'
 
-{CHAR}                   return 'CHAR'
+'/'                                 return 'SLASH'
 
-'[^'                     { this.pushState('class'); return 'NEG_CLASS' }
-'['                      { this.pushState('class'); return 'L_BRACKET' }
+{CHAR}                              return 'CHAR'
+
+'[^'                                { this.pushState(['u','xu'].includes(this.getCurrentState()) ? 'u_class' : 'class'); return 'NEG_CLASS' }
```
DmitrySoshnikov: Let's not allocate an array each time we entry a class. We can do simple `if` check, or allocate it once at the top as a constant, and use here. You can also create a helper function, and reuse it below as well.

DmitrySoshnikov:
CC @mathiasbynens

src/parser/regexp.bnf
Golmote: Ok, changed it.

src/parser/regexp.bnf
```diff
@@ -595,6 +630,9 @@ SourceCharacter
   | ESC_CHAR
     { $$ = Char($1.slice(1), 'simple', @$); $$.escaped = true; }
 
+  | U_CODE_SURROGATE
+    { $$ = Char($1, 'unicode', @$); $$.surrogate_pair = true; }
```
DmitrySoshnikov: let's use JS "camelCase" notation: `isSurrogatePair`

Also, need to update main REAMDE for the `Char` specification for this new field

Golmote: Ok, PR updated!

DmitrySoshnikov:
@Golmote, great job, as always; thanks!

PR #118, opened 2017-11-21
Optimizer: Don't throw on new syntax

Golmote:
Optimizer used to throw an error when used with regexps containing new syntax (like named captures or the `x` flag), since #101 was merged.

It was because I made the Transformer generate an actual RegExp instead of using a string representation. This PR fixes it!


DmitrySoshnikov:
Ah, right, we can't build regexes with non-standard flags; good catch.

PR #119, opened 2017-11-21
Optimizer: Fix (?:abc)+abc optimization in combine-repeating-patterns

Golmote:
I forgot a case in the `increaseQuantifierByOne` function in the `combine-repeating-patterns` transform leading to incorrect optimization of `/(?:abc)+abc/` into `/(?:abc)+/` instead of `/(?:abc){2,}/`.

This PR fixes the case and moves the function to the `utils.js` file since the same behaviour is used in the `quantifiers-merge` transform.

### DIFF
--- a/src/transform/__tests__/transform-utils-test.js
+++ b/src/transform/__tests__/transform-utils-test.js
@@ -43,4 +43,75 @@ describe('transform-utils', () => {
     expect(disjunction).toEqual(expected);
   });
 
+  it('increaseQuantifierByOne', () => {
+    const quantifiers = [{
+        type: 'Quantifier',
+        kind: '*',
+        greedy: true
+    }, {
+      type: 'Quantifier',
+      kind: '+',
+      greedy: false
+    }, {
+      type: 'Quantifier',
+      kind: '?',
+      greedy: true
+    }, {
+      type: 'Quantifier',
+      kind: 'Range',
+      from: 2,
+      to: 2,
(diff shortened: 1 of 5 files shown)


DmitrySoshnikov:
Sounds good!

src/transform/__tests__/transform-utils-test.js
```diff
@@ -43,4 +43,75 @@ describe('transform-utils', () => {
     expect(disjunction).toEqual(expected);
   });
 
+  it('increaseQuantifierByOne', () => {
+    const quantifiers = [{
+        type: 'Quantifier',
+        kind: '*',
+        greedy: true
```
DmitrySoshnikov: BTW, I added recently ESLint checks, you'll need to do `npm install` again to get the latest changes. There should be a pre-commit hook which runs both tests, and ESLint. You can also run it manually as `npm run eslint` in case. Otherwise, publish was blocked by it (https://github.com/DmitrySoshnikov/regexp-tree/commit/4d2bbeeee2414f99702cdeb33eb462cb8ddf19ab).

Golmote: Awesome! I managed to configure the hooks locally and it works like a charm. My PRs should be ok in the future ;)

DmitrySoshnikov: No worries, thanks!

PR #122, opened 2017-11-23
Optimizer: Combine chars sequences into class ranges in char-class-classranges-merge transform

Golmote:
This PR updates the char-class-classranges-merge transform to handle sequences of chars:

`/[abcdef]/` -> `/[a-f]/`

For the sake of readability, I chose to only merge chars being in `\w` or being coded (either unicode, hex, octal or decimal). I'm not sure about this choice though.

Should the optimizer transform `/[<=>?]/` into `/[<-?]/`?

### DIFF
--- a/src/optimizer/transforms/char-class-classranges-merge-transform.js
+++ b/src/optimizer/transforms/char-class-classranges-merge-transform.js
@@ -11,6 +11,7 @@
  * [a-ec] -> [a-e]
  * [a-ec-e] -> [a-e]
  * [\w\da-f] -> [\w]
+ * [abcdef] -> [a-f]
  */
 module.exports = {
   _hasIUFlags: false,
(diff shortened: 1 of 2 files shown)


DmitrySoshnikov:
> Should the optimizer transform /[<=>?]/ into /[<-?]/?

Probably not (yet at least). Even though it might be technically correct, this may look completely differently, and confuse a reader.

src/optimizer/transforms/char-class-classranges-merge-transform.js
```diff
@@ -311,4 +316,53 @@ function fitsInClassRange(expression, classRange) {
...
+      break;
+    }
+  }
+
+  if (nbMergedChars > 1) {
+    expressions[index] = {
+      type: 'ClassRange',
+      from: expressions[index],
+      to: expression
+    };
```
DmitrySoshnikov: Cool 👍

PR #123, opened 2017-11-23
Optimizer: Simplifies class ranges spanning only one or two chars

Golmote:
The transform is this PR handles the following optimizations:

`/[a-a$]/` -> `/[a$]/`
`/[a-b]/` -> `/[ab]/`


DmitrySoshnikov:
Sounds good, thanks!

PR #125, opened 2017-11-24
Parser: When u flag is set, throw error on invalid unicode escape

### LINKED ISSUE #8: `/\e/u` should throw
`/\e/u` is an invalid regular expression: https://mathiasbynens.be/notes/es6-unicode-regex#impact-syntax

So instead of parsing, I’d expect this to throw:

```
$ regexp-tree -e '/\e/u'
{
  "type": "RegExp",
  "body": {
    "type": "Char",
    "value": "e",
    "kind": "simple",
    "escaped": true
  },
  "flags": [
    "u"
  ]
}
```

Golmote:
This should fix #8

### DIFF
--- a/src/parser/__tests__/parser-basic-test.js
+++ b/src/parser/__tests__/parser-basic-test.js
@@ -1068,4 +1068,10 @@ describe('basic', () => {
     });
   });
 
+  it('throws error on invalid Unicode escape', () => {
+    expect(() => parser.parse('/\\p/u')).toThrowError(SyntaxError);
+    expect(() => parser.parse('/\\e/u')).toThrowError(SyntaxError);
+    expect(() => parser.parse('/\\g/u')).toThrowError(SyntaxError);
+  });
+
 });
--- a/src/parser/regexp.bnf
+++ b/src/parser/regexp.bnf
@@ -103,7 +103,14 @@ GROUP_NAME              ([\w$]|\\'u'[0-9a-fA-F]{4}|\\'u{'[0-9a-fA-F]{1,}'}')+
 
 {ESC}'/'                            return 'ESC_CHAR'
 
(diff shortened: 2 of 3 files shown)


src/parser/regexp.bnf
```diff
@@ -103,7 +103,14 @@ GROUP_NAME              ([\w$]|\\'u'[0-9a-fA-F]{4}|\\'u{'[0-9a-fA-F]{1,}'}')+
 
 {ESC}'/'                            return 'ESC_CHAR'
 
-{ESC}{CHAR}                         return 'ESC_CHAR'
+{ESC}{CHAR}                         {
+                                        const s = this.getCurrentState();
+                                        if (s === 'u' || s === 'xu' || s === 'u_class') {
+                                            throw new SyntaxError(`invalid unicode escape ${yytext}`);
```
mathiasbynens: Nit: `s/unicode/Unicode/`

mathiasbynens: If this is only about `\p{…}`, how about: “invalid Unicode property escape”?

Golmote: This is not only about `\p{...}`. All redondant escapes (`\e`, `\a`, `\g`...) will throw this error when `u` flag is set.

src/parser/__tests__/parser-basic-test.js
```diff
@@ -1068,4 +1068,10 @@ describe('basic', () => {
     });
   });
 
+  it('throws error on invalid unicode escape', () => {
```
mathiasbynens: s/unicode/Unicode/

DmitrySoshnikov:
Great! With introducing all `*u` parsing states makes fixing all the Unicode issues in very elegant ways!

PR #126, opened 2017-11-24
Optimizer: Transform Char to lowercase when i flag is set

Golmote:
`/aA[bB]Ï/i` -> `/aa[bb]ï/i`

### DIFF
--- a/src/optimizer/transforms/char-case-insensitive-lowercase-transform.js
+++ b/src/optimizer/transforms/char-case-insensitive-lowercase-transform.js
@@ -1,0 +1,104 @@
+/**
+ * The MIT License (MIT)
+ * Copyright (c) 2017-present Dmitry Soshnikov <dmitry.soshnikov@gmail.com>
+ */
+
+'use strict';
+
+const UPPER_A_CP = 'A'.codePointAt(0);
+const UPPER_Z_CP = 'Z'.codePointAt(0);
+/**
+ * Transforms case-insensitive regexp to lowercase
+ *
+ * /AaBbÏ/i -> /aabbï/i
+ */
+module.exports = {
+  _AZClassRanges: null,
+  _hasUFlag: false,
+  init(ast) {
+    this._AZClassRanges = new Set();
+    this._hasUFlag = ast.flags.includes('u');
+  },
+  shouldRun(ast) {
+    return ast.flags.includes('i');
+  },
+  Char(path) {
+    const {node, parent} = path;
+    if (isNaN(node.codePoint)) {
+      return;
+    }
+
+    // Engine support for case-insensitive matching without the u flag
+    // for characters above \u1000 does not seem reliable.
+    if (!this._hasUFlag && node.codePoint >= 0x1000) {
+      return;
+    }
+
+    if (parent.type === 'ClassRange') {
+      // The only class ranges we handle must be inside A-Z.
+      // After the `from` char is processed, the isAZClassRange test
+      // will be false, so we use a Set to keep track of parents and
+      // process the `to` char.
+      if (!this._AZClassRanges.has(parent) && !isAZClassRange(parent)) {
+        return;
+      }
+      this._AZClassRanges.add(parent);
+    }
+
+    const lower = node.symbol.toLowerCase();
+    if (lower !== node.symbol) {
+      node.value = displaySymbolAsValue(lower, node);
+      node.symbol = lower;
+      node.codePoint = lower.codePointAt(0);
+    }
+  }
+};
+
+function isAZClassRange(classRange) {
+  const {from, to} = classRange;
+  // A-Z
+  return from.codePoint >= UPPER_A_CP && from.codePoint <= UPPER_Z_CP &&
+    to.codePoint >= UPPER_A_CP && to.codePoint <= UPPER_Z_CP;
+}
+
+function displaySymbolAsValue(symbol, node) {
+  const codePoint = symbol.codePointAt(0);
+  if (node.kind === 'decimal') {
+    return '\\' + codePoint;
+  }
+  if (node.kind === 'oct') {
+    return '\\0' + codePoint.toString(8);
+  }
+  if (node.kind === 'hex') {
+    return '\\x' + codePoint.toString(16);
+  }
+  if (node.kind === 'unicode') {
+    if (node.isSurrogatePair) {
+      const {lead, trail} = getSurrogatePairFromCodePoint(codePoint);
+      return '\\u' + '0'.repeat(4 - lead.length) + lead + '\\u' + '0'.repeat(4 - trail.length) + trail;
+    } else if (node.value.includes('{')) {
+      return '\\u{' + codePoint.toString(16) + '}';
+    } else {
+      const code = codePoint.toString(16);
+      return '\\u' + '0'.repeat(4 - code.length) + code;
+    }
+  }
+  // simple
+  return symbol;
+}
+
+/**
+ * Converts a code point to a surrogate pair.
+ * Conversion algorithm is taken from The Unicode Standard 3.0 Section 3.7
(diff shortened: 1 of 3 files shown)


Golmote:
I updated the PR to process more chars.

It now transforms to lowercase all characters it can, but keeps the Char kind.
`/A/i` -> `/a/i`
`/\65/i` -> `/\97/i`
`/\x41/i` -> `/\x61/i`
`/\u0041/i` -> `/\u0061/i`
`/\u{41}/i` -> `/\u{61}/i`

Surrogate pairs should be handled properly as well. I did not include a test, though, since Node 6 does not seem to map lower/upper values of characters above \u{10000}.
Running on Node 7, this transform should work:
`/\ud801\udc09/i` -> `/\ud801\udc31/i`

The transform now also does additional checks if the parent is a Class Range to not break ranges like `A-z`. Only ranges included inside `A-Z` are transformed.

mathiasbynens:
Neat! Maybe add a test like

```js
/\u13AB/ui.test('\uAB7B');
// should be true before and after transformation
```

Golmote:
@mathiasbynens I'm confused. Is the `u` flag required here for `\u13AB` to match `\uAB7B` case-insensitively? It appears to be the case on both NodeJS 6.8, NodeJS 7.4 and Google Chrome. Firefox, though, does not require the `u` flag for it to be true.

mathiasbynens:
I agree it’s confusing. See https://mathiasbynens.be/demo/regex-i and https://github.com/Microsoft/ChakraCore/issues/517 for some background.

Golmote:
@mathiasbynens So, as I understand it, it _is not_ safe to transform `/\u13AB/i` to `/\uAB7B/i` but it _is_ safe to transform `/\u13AB/iu` to `/\uAB7B/iu`? Yet, it is alsways safe to transform `/\u0045/i` to `/\u0065/i`... Do you happen to know where is the limit?

Golmote:
I ran a few tests on my own. Browser support is kind of a mess...
Google Chrome requires the `u` flag for whole ranges of characters (`\u13a0-\u13f5`, `\ua7b2-\ua7b6`...).
Edge 16 fails to lowercase some chars (like `\u10cd` or `\ua69a`), but somehow still matches them properly with `iu` flag.

I updated the PR to require the `u` flag being set before lowercasing characters above `\u1000`.

Even with the `u` flag set, though, the optimisation will still depend on the engine's ability to lowercase the chars. That is, Node 6.8.0 won't optimize `/\u13ab/iu` to `/\uab7b/iu` because it's unable to lowercase `\u13ab` at all.

DmitrySoshnikov:
> Even with the u flag set, though, the optimisation will still depend on the engine's ability to lowercase the chars

Yeah, that's a good decision.

I'm thinking whether we can already bump the engine number to 7 or 8 in the `package.json`. Current stable is 8, and the latest one is already 9.

src/optimizer/transforms/char-case-insensitive-lowercase-transform.js
```diff
@@ -0,0 +1,88 @@
...
+    if (!this._hasUFlag && node.codePoint >= 0x1000) {
+      return;
+    }
+
+    if (parent.type === 'ClassRange') {
+      // The only class ranges we handle are inside A-Z
+      if (isAZClassRange(parent)) {
+        this._AZClassRanges.add(parent);
+      }
+      if (!this._AZClassRanges.has(parent)) {
```
DmitrySoshnikov: So this `if` goes after we do `this._AZClassRanges.add(parent);` on line 38. Does it mean the `return`  should be just in empty `else`?

```js
if (!isAZClassRange(parent)) {
  return;
}

this._AZClassRanges.add(parent);
```

Golmote:
> I'm thinking whether we can already bump the engine number to 7 or 8 in the package.json.

Yeah, that would probably be ok.

src/optimizer/transforms/char-case-insensitive-lowercase-transform.js
```diff
@@ -0,0 +1,88 @@
...
+      node.symbol = lower;
+      node.codePoint = lower.codePointAt(0);
+    }
+  }
+};
+
+function isAZClassRange(classRange) {
+  const {from, to} = classRange;
+  // A-Z
+  return from.codePoint >= 0x41 && from.codePoint <= 0x5a &&
```
DmitrySoshnikov: nit: I always keep forgetting the code points for these, maybe move them to constants above at module level.

```js
const UPPER_A_CP = 'A'.codePointAt(0);

...
// later in function:

return from.codePoint >= UPPER_A_CP ...
```

src/optimizer/transforms/char-case-insensitive-lowercase-transform.js
```diff
@@ -0,0 +1,88 @@
...
+  if (node.kind === 'oct') {
+    return '\\0' + codePoint.toString(8);
+  }
+  if (node.kind === 'hex') {
+    return '\\x' + codePoint.toString(16);
+  }
+  if (node.kind === 'unicode') {
+    if (node.isSurrogatePair) {
+      let trail = 0xdc00 + (codePoint + 0x35fdc00) % 0x400;
+      let lead = (codePoint + 0x35fdc00 - trail) / 0x400;
```
DmitrySoshnikov: These magic number probably good to define as constants as well, or add some comments (maybe to extract to a helper function with a doc-comment, explaining what each number means).

src/optimizer/transforms/char-case-insensitive-lowercase-transform.js
```diff
@@ -0,0 +1,88 @@
...
+      trail = trail.toString(16);
+      lead = lead.toString(16);
+      return '\\u' + '0'.repeat(4 - lead.length) + lead + '\\u' + '0'.repeat(4 - trail.length) + trail;
+    } else if(node.value.includes('{')) {
```
DmitrySoshnikov: space after `if (`

(hm.. wondering why ESLint didn't catch it? maybe it's not configured for this)

src/optimizer/transforms/char-case-insensitive-lowercase-transform.js
Golmote: It's a bit more complicated, since the `isAZClassRange` test will return `false` when we process the `to` char of the range and the `from` char has already been lowercased.

Following your idea, I can still update as:

```js
// The only class ranges we handle must be inside A-Z.
// After the `from` char is processed, the isAZClassRange test
// will be false, so we use a Set to keep track
// of parents and process the `to` char.
if (!this._AZClassRanges.has(parent) && !isAZClassRange(parent)) {
  return;
}
this._AZClassRanges.add(parent);
```

DmitrySoshnikov: Gotcha! OK, sounds good.

src/optimizer/transforms/char-case-insensitive-lowercase-transform.js
Golmote: Good idea, updated!

src/optimizer/transforms/char-case-insensitive-lowercase-transform.js
Golmote: Ok, Google just gave me an easier algorithm, along with an official reference. So even if the numbers look a bit magic, we should be good to go.

src/optimizer/transforms/char-case-insensitive-lowercase-transform.js
Golmote: Indeed, ESLint does not seem to catch it. (I wondered for a minute if my hook was properly set, but even manual `npm run eslint` does not complain for this)

Golmote:
@DmitrySoshnikov Thanks for the thorough review. PR is updated!

DmitrySoshnikov:
Awesome, thanks!

PR #132, opened 2017-11-29
Optimizer: Transform coded chars to simple chars 

Golmote:
The transform in this PR converts coded chars (unicode, hex, oct, dec) to simple chars when they are in the range of ASCII printable chars.

`/\u0041\u0020\u007e/` -> `/A ~/`

It will do the same on class ranges but only when they're included in either `0-9`, `a-z` or `A-Z`.
`/[\x42-\x4c\x6d-\x72]/` -> `/[B-Lm-r]/`

But `/[\u{5a}-\u{61}]/u` is _not_ converted to `/[Z-a]/u`.

It will escape chars when needed (naively, without additional checks):

`/\u005b\u{28}[\u{5c}\x5e]/u` -> `/\[\([\\\^]/u`

This PR relies on (and includes) #130 and #131. Otherwise the tests won't run.
(I'm not sure how annoying it is on your side to handle "PRs with dependencies". If you know of a better way to do this, let me now.)

### DIFF
--- a/src/optimizer/transforms/__tests__/char-code-to-simple-char-transform-test.js
+++ b/src/optimizer/transforms/__tests__/char-code-to-simple-char-transform-test.js
@@ -1,0 +1,116 @@
+/**
+ * The MIT License (MIT)
+ * Copyright (c) 2017-present Dmitry Soshnikov <dmitry.soshnikov@gmail.com>
+ */
+
+'use strict';
+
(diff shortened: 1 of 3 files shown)


DmitrySoshnikov:
Sounds good; I'll take a deeper look later on at this one (BTW, I also merged the dependency PRs, so you can rebase in case).

Issue #129, opened 2017-11-29
Optimizer: Unescape ] char outside CharacterClass


Golmote:
The optimizer could unescape the `]` when it's outside a CharacterClass.

Current behaviour:
`/\][a\]d]\]/` is not optimized.

Expected behaviour:
`/\][a\]d]\]/` -> `/][a\]d]]/`

Golmote:
This issue was fixed with #133.

PR #133, opened 2017-11-30
Optimizer: Unescape ] and fix unescape in class ranges

### LINKED ISSUE #128: Optimizer: char-escape-unescape-transform must handle class ranges
The `char-escape-unescape` transform calls `preservesEscape()` instead of `preservesInCharClass()` for chars inside a class range, because their parent is a ClassRange, not a CharacterClass.

This leads to `/[+-\?]/` not being optimized as `/[+-?]/`.

Golmote:
Fixes #128 and #129


DmitrySoshnikov:
Ah, very good catch! Thanks!

PR #138, opened 2017-12-07
Support Windows for build script. Fix #134

### LINKED ISSUE #134: Support Windows for build script
`npm run build` currently throws errors on Windows.

I usually build manually by calling:
```
node ./node_modules/syntax-cli/bin/syntax -g ./src/parser/regexp.bnf -o ./src/parser/generated/regexp-tree.js -m lalr1 --loc
```

Golmote:
I used `shelljs` as suggested in [your comment](https://github.com/DmitrySoshnikov/regexp-tree/issues/134#issuecomment-348289694) and it seems to work really well!

Note that the symlinks may fail when the command is not run as an administrator.


Golmote:
@DmitrySoshnikov I added the version upgrade for `syntax-cli`

DmitrySoshnikov:
Awesome, thanks!

PR #139, opened 2017-12-07
Optimizer: Don't unescape space and # when x flag is set

### LINKED ISSUE #137: Optimizer: Don't unescape space and # when x flag is set
Currently, an error is thrown when trying to optimize `/\ \#/x` because it unescapes those chars.
I guess the optimizer should handle the `x` flag better to prevent this.

Golmote:
Fix #137


DmitrySoshnikov:
Sounds great!

PR #141, opened 2017-12-11
Optimizer: Don't preserve loc data in transforms. Fix #140

### LINKED ISSUE #140: Optimizer: Must it keep loc values?
I just found out that `regexp-tree -e "/[0123456789]/" -o` throws an error, because [the `char-class-to-meta` transform tries to rewrite the `loc` values](https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/optimizer/transforms/char-class-to-meta-transform.js#L44) while the other transforms (specifically the `char-class-classranges-merge` transform in this case) do not set them.

Are the `loc` values something we want to keep accurate during the transformations?
Even in the `char-class-to-meta` transform, where it's handled, shouldn't it need to rewrite the `loc` values of all nodes following the one that's been replaced?

Golmote:


DmitrySoshnikov:
Yep, thanks!

Issue #135, opened 2017-11-30
Optimizer or minifier ?


Golmote:
This is more of a question than an issue.

Is the Optimizer more an optimizer or a minifier? Or both at the same time?

Currently, the Optimizer does some "optimizations" that actually increase the length of the regexp, like:
`/aa/` (2) -> `/a{2}/` (4)
`/a|b/` (3) -> `/[ab]/` (4)
`/aa+/` (3) -> `/a{2,}/` (5)
If it's supposed to be a minifier, we should try to avoid those cases. (Note that this could be tricky sometimes, since an optimization might not seem worth it at first, but work well with following optimizations)

Also, if we really want minification, we could add additional transforms like:
find the shortest representation of a char:
`/\u0010/` (6) -> `/\x10/` (4)
`/\u0000/` (6) -> `/\0/` (2)
convert all coded chars to their printable equivalent (even outside of basic printable ASCII range):
`/\u{dd}/u` (6) -> `/Ý/` (1)
reorder class ranges in a way to avoid escaping:
`/[\^a]/` (5) -> `/[a^]/` (4)
`/[,\-.]/` (6) -> `/[,.-]/` (5)
and so on.

What's your opinion on this?

I intend to start working on actual options for the optimizer. The whitelist is a good start, but it's kind of verbose and does not really allow for "optional" transforms. I think an object of options with default values would be more appropriate (something like the [compress options in UglifyJS](https://github.com/mishoo/UglifyJS2#compress-options)).

DmitrySoshnikov:
@Golmote, all good points. Initially I was thinking in terms of "minifier", however, "optimizer" might also mean whether it's faster to parse/execute a specific construct.

The things with printable chars, and making codes simpler also makes sense, would be great to have.

Potentially we can provide also `--minimize`, unless it'll be confusing what to use in which use-case: `--optimize` or `--minimize`. So we should keep in mind that it should be easier for an end user (making less decisions/choice in this case). At the same time, I like the idea of passing options, and using specific transforms for advanced users.

Golmote:
@DmitrySoshnikov I think the default options should lead to an output that follows those principles:

- Optimize and minimize as much as possible
- Don't (ever?) make the optimized regexp longer than the original
- Maintain readability

(By readability, I mean avoiding conversion to unicode chars, avoid ranges that don't "feel" right like `Z-a`, etc.)

That should keep the standard end user happy. And options could provide a more granular approach.

[...]

PR #151, opened 2018-03-01
Optimizer: Don't apply transform if result is longer

Golmote:
Related to #135 and more specifically https://github.com/DmitrySoshnikov/regexp-tree/issues/135#issuecomment-366053266, this PR prevents a transform from being applied if the result is longer that the original.

### DIFF
--- a/src/optimizer/index.js
+++ b/src/optimizer/index.js
@@ -32,9 +32,17 @@ module.exports = {
       ? transformsWhitelist
       : Object.keys(optimizationTransforms);
 
-    let result = new transform.TransformResult(parser.parse(regexp));
+    let ast = regexp;
+    if (regexp instanceof RegExp) {
+      regexp = `${regexp}`;
+    }
+
+    if (typeof regexp === 'string') {
+      ast = parser.parse(regexp);
+    }
+
+    let result = new transform.TransformResult(ast);
     let prevResult;
-    let ast;
 
     do {
       prevResult = result.toString();
--- a/src/optimizer/index.js
+++ b/src/optimizer/index.js
@@ -5,7 +5,7 @@
 
 'use strict';
 
-const clone = require('clone');
+const clone = require('../utils/clone');
 const parser = require('../parser');
 const transform = require('../transform');
 const optimizationTransforms = require('./transforms');
--- a/src/optimizer/index.js
+++ b/src/optimizer/index.js
@@ -5,6 +5,8 @@
 
 'use strict';
 
+const clone = require('clone');
+const parser = require('../parser');
 const transform = require('../transform');
 const optimizationTransforms = require('./transforms');
 
@@ -30,13 +32,14 @@ module.exports = {
       ? transformsWhitelist
       : Object.keys(optimizationTransforms);
 
+    let result = new transform.TransformResult(parser.parse(regexp));
     let prevResult;
-    let result;
+    let ast;
+
     do {
-      if (result) {
-        prevResult = result.toString();
-        regexp = prevResult;
-      }
+      prevResult = result.toString();
+      ast = clone(result.getAST());
+
       transformToApply.forEach(transformName => {
 
         if (!optimizationTransforms.hasOwnProperty(transformName)) {
@@ -49,9 +52,13 @@ module.exports = {
 
         const transformer = optimizationTransforms[transformName];
 
-        result = transform.transform(regexp, transformer);
-        regexp = result.getAST();
+        let newResult = transform.transform(ast, transformer);
+        if (newResult.toString().length <= result.toString().length) {
+          result = newResult;
+        }
+        ast = clone(result.getAST());
       });
+
     } while (result.toString() !== prevResult);
 
     return result;
--- a/src/optimizer/index.js
+++ b/src/optimizer/index.js
@@ -42,10 +42,13 @@ module.exports = {
     }
 
     let result = new transform.TransformResult(ast);
-    let prevResult;
+    let prevResultString;
 
     do {
-      prevResult = result.toString();
+      // Get a copy of the current state here so
+      // we can compare it with the state at the
+      // end of the loop.
+      prevResultString = result.toString();
       ast = clone(result.getAST());
 
       transformToApply.forEach(transformName => {
@@ -60,19 +63,24 @@ module.exports = {
 
         const transformer = optimizationTransforms[transformName];
 
+        // Don't override result just yet since we
+        // might want to rollback the transform
         let newResult = transform.transform(ast, transformer);
 
         if (newResult.toString() !== result.toString()) {
           if (newResult.toString().length <= result.toString().length) {
             result = newResult;
           } else {
-            // Result has changed but is not shorter: restore ast to its previous state.
+            // Result has changed but is not shorter:
+            // restore ast to its previous state.
             ast = clone(result.getAST());
           }
         }
       });
 
-    } while (result.toString() !== prevResult);
+      // Keep running the optimizer until it stops
+      // making any change to the regexp.
+    } while (result.toString() !== prevResultString);
 
     return result;
   },
--- a/src/optimizer/index.js
+++ b/src/optimizer/index.js
@@ -53,10 +53,15 @@ module.exports = {
         const transformer = optimizationTransforms[transformName];
 
         let newResult = transform.transform(ast, transformer);
-        if (newResult.toString().length <= result.toString().length) {
-          result = newResult;
+
+        if (newResult.toString() !== result.toString()) {
+          if (newResult.toString().length <= result.toString().length) {
+            result = newResult;
+          } else {
+            // Result has changed but is not shorter: restore ast to its previous state.
+            ast = clone(result.getAST());
+          }
         }
-        ast = clone(result.getAST());
       });
 
     } while (result.toString() !== prevResult);
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,9 +55,9 @@
       }
     },
     "ajv-keywords": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-2.1.1.tgz",
-      "integrity": "sha1-YXmX/F9gV2iUxDX5QNgZ4TW4B2I=",
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-3.1.0.tgz",
+      "integrity": "sha1-rCsnk5xUPpXSwG5/f1wnvkqlQ74=",
       "dev": true
     },
     "align-text": {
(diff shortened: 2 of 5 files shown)


src/optimizer/index.js
```diff
@@ -30,13 +32,14 @@ module.exports = {
...
-    let result;
+    let ast;
+
     do {
-      if (result) {
-        prevResult = result.toString();
-        regexp = prevResult;
-      }
+      prevResult = result.toString();
+      ast = clone(result.getAST());
```
DmitrySoshnikov: We'll need to check for speed regressions. When the optimizer is used on a huge cobe base, this might be a potential performance regression. Do you know how fast/slow this `clone` is?

src/optimizer/index.js
```diff
@@ -49,9 +52,13 @@ module.exports = {
 
         const transformer = optimizationTransforms[transformName];
 
-        result = transform.transform(regexp, transformer);
-        regexp = result.getAST();
+        let newResult = transform.transform(ast, transformer);
+        if (newResult.toString().length <= result.toString().length) {
+          result = newResult;
+        }
+        ast = clone(result.getAST());
```
DmitrySoshnikov: do we do `clone` even twice within this `do-while` loop?

src/optimizer/index.js
Golmote: I don't, no. 
The thing is I can't rollback a transform if I can't restore the AST to its previous state. And since `transform()` mutates the AST, I do need a copy here.

I might try to make some kind of benchmark with different versions of regexp-tree.

Golmote:
@DmitrySoshnikov I added a condition so that the AST is cloned less often: once at the beginning and then only when we need to rollback the transform.

Golmote:
Quick'n'dirty benchmark:
I ran `regexpTree.optimize()` 1000 times for each regexp and simply averaged the execution time.

Regexp | `master` time (ms) | `master` result | `shorter-optimisations` time (ms) | `shorter-optimisations` result
---|---|---|---|---
`/[\r\n\t ][\d\w\W]/` | 0.547 | `/\s[\W\w]/` | 0.564 | `/\s[\W\w]/`
`/aa[\d\w\W]/` | 0.462 | `/a{2}[\W\w]/`  | 0.629 | `/aa[\W\w]/`
`/(?:abc)(?:abc)aa[\d\w\W]/` | 0.932 | `/abc{2}a{2}[\W\w]/`  | 1.263 | `/abcabcaa[\W\w]/`
`/(?:abcde)(?:abcde)aa[\d\w\W]/` | 1.169 | `/abcde{2}a{2}[\W\w]/`  | 1.083 | `/abcde{2}a{2}[\W\w]/`
`/(?:a\|b\|c\|d)[\d\w\W]/` | 0.95 | `/[a-d][\W\w]/`  | 0.874 | `/[a-d][\W\w]/`
`/(?:a\|b)[\d\w\W]/` | 0.476 | `/[ab][\W\w]/`  | 0.46 | `/[ab][\W\w]/`
`/aa+[\d\w\W]/` | 0.459 | `/a{2,}[\W\w]/`  | 0.674 | `/aa+[\W\w]/`

src/optimizer/index.js
```diff
@@ -30,13 +32,14 @@ module.exports = {
       ? transformsWhitelist
       : Object.keys(optimizationTransforms);
 
+    let result = new transform.TransformResult(parser.parse(regexp));
```
DmitrySoshnikov: The `optimize` potentially can accept the AST directly, which won't work with the `parse.parse`; we should consider this use-case here.

src/optimizer/index.js
```diff
@@ -49,9 +52,18 @@ module.exports = {
...
-        result = transform.transform(regexp, transformer);
-        regexp = result.getAST();
+        let newResult = transform.transform(ast, transformer);
+
+        if (newResult.toString() !== result.toString()) {
+          if (newResult.toString().length <= result.toString().length) {
+            result = newResult;
+          } else {
+            // Result has changed but is not shorter: restore ast to its previous state.
+            ast = clone(result.getAST());
```
DmitrySoshnikov: So it seems to me, we can just re-parse in this case (since we already were doing this in the main `do-while` loop, previous line 38, where `regexp = prevResult;`, and which was the string).

The `clone` is a very generic clone function, which considers several edges cases (like cyclic references, different types of objects, etc). Plus, we're introducing another dependency.

I think we can try just doing here the:

```js
regexp = result.toString();
```

and it can normally be fed further to the `transform.transform(...)`. Re-parsing is specific to this structure, and potentially might be faster than the generic clone on larger code-bases.

I haven't tested it yet, but how about this:

```js
optimize(regexp, transformsWhitelist = []) {
    const transformToApply = transformsWhitelist.length > 0
      ? transformsWhitelist
      : Object.keys(optimizationTransforms);

    let prevResult = new transform.TransformResult(
      typeof regexp === 'object' ? regexp
      : parser.parse(`${regexp}`);
    );

    let result;

    do {
      transformToApply.forEach(transformName => {

        if (!optimizationTransforms.hasOwnProperty(transformName)) {
          throw new Error(
            `Unknown optimization-transform: ${transformName}. ` +
            `Available transforms are: ` +
            Object.keys(optimizationTransforms).join(', ')
          );
        }

        const transformer = optimizationTransforms[transformName];

        // Obtain the next result.
        result = transform.transform(regexp, transformer);

        // New transformation resulted to a longer regexp, restore:
        if (prevResult.toString().length < result.toString().length) {
          regexp = prevResult.toString();
        } else {
          // Otherwise, proceed with the same AST:
          regexp = result.getAST();
        }

        prevResult = result;
      });
    } while (result.toString() !== prevResult.toString);

    return result;
  },
```

Golmote: @DmitrySoshnikov Ok, so I gave it a try, basically replacing the call to `clone()` with `result.toString()`. This does not help at all with the performance :cry: 
All my previous tests take 2 to 3 times longer to run!

I think parsing the regexp takes much more time than cloning the AST.

Regarding the new dependency and the unneeded complexity, would you be ok with a custom and simple deep copy implementation?

DmitrySoshnikov: Hm.. interesting. Yeah, if the `clone` implemented correctly, it might be better, and in this case we can introduce this dependency I think. OK, sounds good, I'll take another look at the evening, and will merge in case. Thanks for the tests!

Golmote: I just noticed the clone library I picked in this PR has a second parameter than can be set to `false` to disable the handling of circular data to improve performance.

> This function supports circular references by default, but if you are certain
> there are no circular references in your object, you can save some CPU time
> by calling clone(obj, false).
> Caution: if `circular` is false and `parent` contains circular references,
> your program may enter an infinite loop and crash.

Maybe we can take advantage of this, since the AST has no reason to have circular references, right?

DmitrySoshnikov: Yeah, also, that's right. Also, the only JS objects which can appear in AST are `{}`, and `[]`, i.e. not dates, functions, etc. So from this perspective, a custom and simple clone utility as you mentioned might a good alternative actually, since we will not need to handle all those edge case.

DmitrySoshnikov: If we go with the custom clone util, we can put it in `src/utils/clone.js`. And we can just take basic functionality from the `clone` package, and remove all edges cases making it a simple for the use-case.

Golmote:
New version with a custom clone util.

Updated benchmark (I also reran the `master` ones just to be sure), it's slightly improved compared to the version with dependency.

Regexp | `master` time (ms) | `master` result | `shorter-optimisations` time (ms) | `shorter-optimisations` result
---|---|---|---|---
`/[\r\n\t ][\d\w\W]/` | 0.536 | `/\s[\W\w]/` | 0.511 | `/\s[\W\w]/`
`/aa[\d\w\W]/` | 0.464 | `/a{2}[\W\w]/`  | 0.49 | `/aa[\W\w]/`
`/(?:abc)(?:abc)aa[\d\w\W]/` | 0.933 | `/abc{2}a{2}[\W\w]/`  | 0.994 | `/abcabcaa[\W\w]/`
`/(?:abcde)(?:abcde)aa[\d\w\W]/` | 1.137 | `/abcde{2}a{2}[\W\w]/`  | 1.038 | `/abcde{2}a{2}[\W\w]/`
`/(?:a\|b\|c\|d)[\d\w\W]/` | 0.941 | `/[a-d][\W\w]/`  | 0.792 | `/[a-d][\W\w]/`
`/(?:a\|b)[\d\w\W]/` | 0.484 | `/[ab][\W\w]/`  | 0.427 | `/[ab][\W\w]/`
`/aa+[\d\w\W]/` | 0.463 | `/a{2,}[\W\w]/`  | 0.503 | `/aa+[\W\w]/`

DmitrySoshnikov:
@Golmote, looks good overall! Seems though we don't need 3 variables (`result`, `newResult`, `prevResult`), and can use only 2, just maintaining needed ast.

How about this?

```js
optimize(regexp, transformsWhitelist = []) {
  const transformsToApply = transformsWhitelist.length > 0
    ? transformsWhitelist
    : Object.keys(optimizationTransforms);

  let ast = regexp;
  if (regexp instanceof RegExp) {
    regexp = `${regexp}`;
  }

  if (typeof regexp === 'string') {
    ast = parser.parse(regexp);
  }

  let prevResult;
  let result = new transform.TransformResult(ast);

  do {
    prevResult = result;

    transformsToApply.forEach(transformName => {
      const transformer = optimizationTransforms[transformName];
      result = transform.transform(ast, transformer);

      if (result.toString() !== prevResult.toString()) {
        if (result.toString().length <= prevResult.toString().length) {
          ast = result.getAST();
        } else {
          // Result has changed but is not shorter:
          // restore ast to its previous state.
          ast = clone(prevResult.getAST());
        }
      }
    });

  } while (result.toString() !== prevResult.toString());

  return result;
},
```

Golmote:
@DmitrySoshnikov I thought about this when you posted https://github.com/DmitrySoshnikov/regexp-tree/pull/151#discussion_r172288351, but I'm pretty sure we do need three here.

The `do/while` loop is used to ensure we keep running the optimizer until it can't optimize anything more, so we need to be able to compare the state at the very beginning of the loop (`prevResult`) with the state at the very end of the loop after all transforms have run (`result`).

But in the `forEach` loop, we need to be able to revert _only the last_ transform that occurred. That's why we need to compare the state just before the last transform (`result`) with the state just after the last transform (`newResult`).

Do you think I should add comments to make this more clear?

DmitrySoshnikov:
@Golmote, OK, sounds good, and yeah, some explanation comments would be good, so we won't forget later the logic there. And then we can merge it.

Golmote:
@DmitrySoshnikov I hope my comments make it more clear now.

DmitrySoshnikov:
OK, looks good, thanks, @Golmote!

PR #153, opened 2018-03-04
Optimizer: Ungroup: Merge Group content of type Alternative with parent Alternative

### LINKED ISSUE #152: Optimizer: Ungroup transform does not merge adjacent Alternatives
At the moment, when the Ungroup transform removes groups, it does not merge adjacent Alternatives, leading to an AST that should not naturally happen on parsing.

For example: `(?:abcd)(?:abcd)` is an Alternative containing two Groups.
When the Ungroup transform runs, it transforms it to an Alternative of two Alternatives. This gives the correct result when generated as a string or a regexp, but the AST should not contain this.

This issue leads to following transforms misinterpreting the AST:
`(?:abcd)(?:abcd)` -> `abcdabcd` -> `abcd{2}` instead of `(?:abcd){2}`.

Golmote:
Fixes #152.

### DIFF
--- a/src/optimizer/transforms/__tests__/ungroup-transform-test.js
+++ b/src/optimizer/transforms/__tests__/ungroup-transform-test.js
@@ -50,6 +50,20 @@ describe('ungroup', () => {
     expect(re.toString()).toBe('/[ab]+/');
   });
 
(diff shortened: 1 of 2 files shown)


DmitrySoshnikov:
Yeah, pretty subtle issue. Thanks!

PR #157, opened 2018-03-12
Optimizer: CharClassClassrangesMerge: Only merge \w chars and char codes into class ranges

### LINKED ISSUE #156: Wrong classranges optimization
Something is wrong with classranges optimizations, example here:

    $ bin/regexp-tree -e '/[^a-zA-Z0-9_!#$%&*@＠]/' -o
   
    Optimized: /[^\d!#$%&*@-Z_a-z＠]

With `charClassClassrangesMerge` optimization disabled I have better result:

    Expected:  /[^\w!#$%&*@＠]/

Golmote:
This fixes #156.

EDIT: Somehow I did not get notifications for the discussion that happened in #156 between yesterday evening and this morning. I still think this PR might be relevant though, since we already had a special case to only merge word chars and char codes in a class range [in some cases](https://github.com/DmitrySoshnikov/regexp-tree/pull/157/files#diff-48e74382c584f8474d8aadcdb8af4beeR162).

### DIFF
--- a/src/optimizer/transforms/__tests__/char-class-classranges-merge-transform-test.js
+++ b/src/optimizer/transforms/__tests__/char-class-classranges-merge-transform-test.js
@@ -73,7 +73,7 @@ describe('char-class-classranges-merge', () => {
     expect(re.toString()).toBe('/[\\S]/');
   });
 
-  it('merges chars in class ranges', () => {
+  it('merges \\w chars and char codes in class ranges', () => {
     let re = transform(/[fb-eg]/, [
       charClassClassrangesMerge,
     ]);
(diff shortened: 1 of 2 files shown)


src/optimizer/transforms/__tests__/char-class-classranges-merge-transform-test.js
```diff
@@ -100,6 +100,13 @@ describe('char-class-classranges-merge', () => {
     expect(re.toString()).toBe('/[\\ud83d\\ude80-\\ud83d\\ude88]/u');
   });
 
+  it('does not merge non \\w chars into class ranges to keep readability', () => {
```
silentroach: but why? optimization isn't about readability

Golmote: At the moment, the optimizer tries to find the right balance between optimization, strictness and readability. Once we add options, we'll be able to have more atomic settings.

DmitrySoshnikov: Yeah, I think it's a good trade-off, thanks guys!

davisjam:
+1

Issue #154, opened 2018-03-07
Interested in more examples?


davisjam:
This is just a question.

I've written two small scripts based on the traversal API.

1. Script 1 implements substack's [safe-regex](https://github.com/substack/safe-regex) project, but it fixes the known bugs in that program.
2. Script 2 checks for the presence of certain assertion types in a regex.

Both use pre-processing that converts (some) Python syntax into regexp-tree syntax.

@DmitrySoshnikov Would you like these scripts in `examples/`?

DmitrySoshnikov:
@davisjam, yep, good idea, the more good practical examples, the better!

Re: `safe-regex`, I see it uses `ret` package as a dependency, which is a "regexp tokenizer", and can be limited by not having full AST. If you fix some issues in that program, this is awesome!

davisjam:
I have submitted PRs to `safe-regex`. The author appears to have abandoned it.

`safe-regex` is actually how I found `regexp-tree` -- dissatisfaction with `ret` led me here.

davisjam:
> Yep, good idea, the more good practical examples, the better!

OK. I'll open a PR in a few days.

davisjam:
@DmitrySoshnikov FYI I now have publish rights to `safe-regex` and am working on a new version that uses `regexp-tree` and embeds the examples we talked about in #159.

DmitrySoshnikov:
@davisjam, this sounds great! Let me know in case you'll need any support or help.

PR #159, opened 2018-03-12
examples: star height

davisjam:
Add an implementation of the "star height" heuristic
to detect regexes vulnerable to REDOS.

See also [https://github.com/substack/safe-regex](safe-regex).

Towards #154.


davisjam:
@DmitrySoshnikov Not quite in the style of the rest of the module. Let me know your thoughts.

examples/count-star-height.js
```diff
@@ -0,0 +1,31 @@
...
+
+const file = process.argv[2];
+const content = fs.readFileSync(file);
+const obj = JSON.parse(content);
+
+try {
+	const starHeight = countStarHeight(obj.pattern, obj.options);
+	obj.starHeight = starHeight;
+}
+catch (e) {
```
DmitrySoshnikov: You can install Prettier for auto-formatting (I'll open an issue with the default config file for it). Overall, JS style is:

```js
try {
  ...
} catch (e) {
  ...
}
```

examples/count-star-height.js
```diff
@@ -0,0 +1,31 @@
...
+  console.error(e);
+	obj.starHeight = 'UNKNOWN';
+}
+
+console.log(JSON.stringify(obj));
```
DmitrySoshnikov:
```js
JSON.stringify(obj, null, 2);
```

examples/lib/count-star-height.js
```diff
@@ -0,0 +1,83 @@
+/* Author: Jamie Davis <davisjam@vt.edu>
+ * Description: Lib to count the star height of a regex.
+ *              This is a correct and improved version of substack's safe-regex.
+ *
+ * Star height is a heuristic for REDOS (exponential blow-up).
+ */
+
+const createRegExp = require('./create-regexp'),
+      regexpTree = require('../../');
```
DmitrySoshnikov: ditto, please use standard style in the repo

examples/lib/count-star-height.js
```diff
@@ -0,0 +1,83 @@
+/* Author: Jamie Davis <davisjam@vt.edu>
```
DmitrySoshnikov: ditto, please use standard doc header with license, and copyright (and you can keep your name if you'd like)

examples/lib/count-star-height.js
```diff
@@ -0,0 +1,83 @@
...
+ *              default false (i.e. ignore ?'s)
+ *            minimumRepetitionUpperLimit -- a{0,3} is probably OK, a{0,50} is not.
+ *              default 0 (i.e. all of {x}, {x,}, {,x}, {x,y} are problematic)
+ *
+ * Returns the regex's star height: an integer >= 0.
+ */
+function countStarHeight(regex, options) {
+	/* Get an ast. */
+	const re = createRegExp(regex);
+	console.error(`re: /${re.source}/`);
```
DmitrySoshnikov: why is it `console.error`?

can use `console.info(...)` for normal messages

examples/lib/count-star-height.js
```diff
@@ -0,0 +1,83 @@
...
+	const ast = regexpTree.parse(re);
+
+	/* Options. */
+	const default_countQuestionMarks = false;
```
DmitrySoshnikov: JS style, camelCase: `defaultCountQuestionMarks`

examples/lib/count-star-height.js
```diff
@@ -0,0 +1,83 @@
...
+	const default_minimumRepetitionUpperLimit = 0;
```
DmitrySoshnikov: ditto

examples/lib/count-star-height.js
```diff
@@ -0,0 +1,83 @@
...
+		options = {};
+	}
+	// Apply defaults.
+	if (!options.countQuestionMarks) {
+		options.countQuestionMarks = default_countQuestionMarks;
+	}
+	if (!options.minimumRepetitionUpperLimit) {
+		options.minimumRepetitionUpperLimit = default_minimumRepetitionUpperLimit;
+	}
+	console.error(`countQuestionMarks ${options.countQuestionMarks} minimumRepetitionUpperLimit ${options.minimumRepetitionUpperLimit}`);
```
DmitrySoshnikov: `console.info`

examples/lib/count-star-height.js
```diff
@@ -0,0 +1,83 @@
...
+
+	/* Here we go! */
+
+	let currentStarHeight = 0;
+	let maxObservedStarHeight = 0;
+
+	regexpTree.traverse(ast, {
+		'Repetition': {
```
DmitrySoshnikov: no quotes just `Repetition:`

examples/lib/count-star-height.js
```diff
@@ -0,0 +1,83 @@
...
+			pre({node}) {
+				// Optional things to ignore
+				if (node.quantifier && node.quantifier.kind === '?' && !options.countQuestionMarks) {
+					return;
+				}
+				else if (node.quantifier.kind === 'Range' &&
```
DmitrySoshnikov: on previous line

examples/lib/count-star-height.js
```diff
@@ -0,0 +1,83 @@
...
+				currentStarHeight++;
+				if (maxObservedStarHeight < currentStarHeight) {
+					maxObservedStarHeight = currentStarHeight;
+				}
+			},
+			post({node}) {
+				if (node.quantifier && node.quantifier.kind === '?' && !options.countQuestionMarks) {
+					return;
+				}
+				else if (node.quantifier.kind === 'Range' &&
```
DmitrySoshnikov:
```js
if (...) {
  ...
} else if (...) {
  ...
}
```

examples/lib/create-regexp.js
```diff
@@ -0,0 +1,75 @@
+const regexpTree = require('../../');
```
DmitrySoshnikov: Use standard doc block comment with license and copyright.

examples/lib/create-regexp.js
```diff
@@ -0,0 +1,75 @@
...
+
+  /* Convert Python syntax to the regexpTree regex syntax. */
+
+	// Named capture groups.
+	pattern = pattern.replace(/\(\?P(<(?:.*?)>.*?)\)/g, function(str, group1) {
+		return `(?${group1})`;
+	});
+
+	// Remove comments.
+	pattern = pattern.replace(/\(\?#.*?\)/g, function(str, group1) {
```
DmitrySoshnikov: use arrow function 

```js
..., (str, group1) => {
```

examples/lib/create-regexp.js
```diff
@@ -0,0 +1,75 @@
...
+		return '';
+	});
+
+	let reOrPattern;
+	try {
+		reOrPattern = new RegExp(pattern);
+	}
+	catch (e) {
```
DmitrySoshnikov: style

examples/count-star-height.js
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
davisjam: If you want to run it as `./count-star-height.js` without `node ./count-star-height.js` they are. For my purposes this was preferable, but if you would prefer the example need be invoked with the Node interpreter then I can switch it.

examples/count-star-height.js
```diff
@@ -0,0 +1,31 @@
...
+/* Author: Jamie Davis <davisjam@vt.edu>
```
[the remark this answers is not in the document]
davisjam: OK

examples/count-star-height.js
davisjam: I'll try out that tool.

examples/lib/count-star-height.js
davisjam: For my use of this example I wanted stdout to only contain "the answer". Any progress updates were to stderr to make it easier for me to script around the program. I can switch to `console.info` though.

davisjam:
Thoughts on including the linter as part of your git commit and push hooks? I assumed that was already in place.

DmitrySoshnikov:
Yeah, just an FYI, in https://github.com/DmitrySoshnikov/regexp-tree/commit/611adc23103665bdbbfc292982368bb4268afcca and https://github.com/DmitrySoshnikov/regexp-tree/commit/140938592b3f28ab1f05c1c8879764963fa83d88 I have setup Prettier. So if you rebase, do `npm install`, and `npm run build`, the linter, and Prettier formatter should run as a pre-commit hook. You can also setup the [pre-commit](https://github.com/DmitrySoshnikov/regexp-tree/blob/master/scripts/git-pre-commit) hook manually.

davisjam:
@DmitrySoshnikov I believe I've addressed your review.


As I noted above, I'm leery about giving you copyright since this example is excerpted from my current research project. Unless I misunderstand what "copyright" means in this context?

examples/lib/count-star-height.js
```diff
@@ -0,0 +1,85 @@
...
+  let maxObservedStarHeight = 0;
+
+  regexpTree.traverse(ast, {
+    'Repetition': {
+      pre({node}) {
+        // Optional things to ignore
+        if (node.quantifier && node.quantifier.kind === '?' && !options.countQuestionMarks) {
+          return;
+        } else if (node.quantifier.kind === 'Range' &&
+                   (!('to' in node.quantifier) || node.quantifier.to < options.minimumRepetitionUpperLimit))
```
DmitrySoshnikov: Let's run Prettier nevertheless, it'll reformat it in the single style.

examples/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Here you'll find ways people have used regexp-tree for fun and profit.
+
+Each example has:
+- A command-line interface
+- A corresponding `lib/` entry for ease of adoption
+
+| File                   | Author                                               | Description       |
+|------------------------|------------------------------------------------------|-------------------|
+| `count-star-height.js` | Jamie Davis [@davisjam](https://github.com/davisjam) | Correct implementation of the well-known and buggy [safe-regex](https://github.com/substack/safe-regex) project. |
```
DmitrySoshnikov: OK, here what I though.

This actually might be a good separate project by itself. Maybe you should create a separate repo for this at all instead of having it in the main regexp-tree repo? With this you fully preserve your own copyright.

Since the example is specific, it might be less practical as specifically an example for a quick start to use the tool. However, I can see it as a good example of already using the tool in other good projects like your research project.

davisjam: OK, so there are really two discussions here:

1. Whether you're interested in `examples/` (and if so, what kind?)
2. Copyright.

I'm certainly not trying to start a fight about this, I just want to make sure we're on the same page.

# To `examples/` or not to `examples/`

> This actually might be a good separate project by itself.

Once the paper is accepted, I will be publishing my research repo to GitHub. The repo includes  this use of `regexp-tree` and a few others, plus a bunch of other unrelated code.

However, I'm loath to just dump my code on the internet and forget about it. I thought this would be a good place to put the `regexp-tree`-specific scripts as a way of giving back to the project.

We could link to a separate project with examples in the README, but then you end up with a link for every person who wants to share examples. Centralization seems better to me in this case.

> Since the example is specific

Of course it's specific :-). The entire `regexp-tree` project is pretty niche. Awesome, but niche.

> it might be less practical as specifically an example for a quick start to use the tool.

There are already many "quick-start" examples in the main README that demonstrate how to use individual features. I view the purpose of `examples/` as to show more complex uses of `regexp-tree` so people can get a sense of how they might want to use it themselves.

Did you have a different kind of code in mind for `examples/`?

# Copyright

> With this you fully preserve your own copyright.

If you want the examples it seems like we can address this with appropriate amendments to the copyright line in a few files.

DmitrySoshnikov: @davisjam, yeah, that's correct, I'd like regexp-tree to maintain base core functionality, don't overload it with "concrete sub-projects" (like your research project), and avoid explicitly promoting/depending on other projects with direct links in README, etc.

All the contributors to this specific repo are explicitly mentioned in the https://github.com/DmitrySoshnikov/regexp-tree/graphs/contributors, with their explicit commits, etc, so their "copyright" is automatically "preserved" (assuming the MIT License of this project).

With this, the `examples\` directory I think should be completely independent from other projects (even as an "example dependency"), less focus on individual names, pages, etc, and more focus on actual examples which can help start using regexp-tree for some needs. These examples, as it seems to me, can be in general still very simple, but taking some smaller "real-world" issues.

For the specific research project, I really think it'll be better to open a separate repo. This will be much better for it, instead of putting it into some generic repo (which regexp-tree is).

davisjam:
Acknowledged. I've petitioned npm for publish rights to [safe-regex](https://github.com/substack/safe-regex) and will be putting this code there.

DmitrySoshnikov:
Sounds good, thanks!

Issue #164, opened 2018-10-28
Nested capture group numbering is inverted


davisjam:
The regex `/(((a)b)c)/` contains 3 groups. Nested groups are numbered left-to-right by left parentheses. So backreference `\3` matches "a", backreference `\2` matches "ab", and backreference `\1` matches "abc".

For example:

```
(14:43:51) jamie@woody ~/Desktop/EcosystemRegexps/analyze-regexps/structural-analysis $ node
> /(((a)b)c)\1/.exec('abcabc')
[ 'abcabc',
  'abc',
  'ab',
  'a',
  index: 0,
  input: 'abcabc',
  groups: undefined ]
> /(((a)b)c)\2/.exec('abcab')
[ 'abcab',
  'abc',
  'ab',
  'a',
  index: 0,
  input: 'abcab',
  groups: undefined ]
> /(((a)b)c)\3/.exec('abca')
[ 'abca',
  'abc',
  'ab',
  'a',
  index: 0,
  input: 'abca',
  groups: undefined ]
> 
```

I am running regexp-tree v0.0.85.

It looks to me like nested groups are numbered in reverse, with the outermost given the largest number and the innermost given the smallest.

```
re: /(((a*)))\1/
pattern /(((a*)))\1/ yields AST:
{
  "type": "RegExp",
  "body": {
    "type": "Alternative",
    "expressions": [
      {
        "type": "Group",
        "capturing": true,
        "number": 3,
        "expression": {
          "type": "Group",
          "capturing": true,
          "number": 2,
          "expression": {
            "type": "Group",
            "capturing": true,
            "number": 1,
            "expression": {
              "type": "Repetition",
              "expression": {
                "type": "Char",
                "value": "a",
                "kind": "simple",
                "symbol": "a",
                "codePoint": 97
              },
              "quantifier": {
                "type": "Quantifier",
                "kind": "*",
                "greedy": true
              }
            }
          }
        }
      },
      {
        "type": "Backreference",
        "kind": "number",
        "number": 1,
        "reference": 1
      }
    ]
  },
  "flags": ""
}
```

davisjam:
I believe this is due to the recursive nature of the parsing.

regexp.bnf includes:

```
Capturing Group: ... | L_PAREN Disjunction R_PAREN
{
  capturingGroupsCount++
  // Create Node
}
```

Because we resolve the language recursively (I assume?), we update `capturingGroupsCount` and create the `Node` for inner groups before outer groups.

[...]

PR #165, opened 2018-10-29
WIP: Number nested groups correctly

davisjam:
Towards #164.

1. Possibly over-engineering. Not sure if there's a need for a "PostParsingPass" concept. But I think it's cleaner.

2. As noted in 5d3682b, I'm not sure how to construct the list of nodes in a way that enables the re-numbering we want, *without* traversing the full tree.

```
=======================
( a ( b ( c ) ) ) d ( e)

Visit order:
3   2   1              4

Desired number order:
1   2   3              4

=======================
( a ) ( b ) ( c ) ( d )

Visit order:
1     2       3      4

Desired number order:
1     2       3      4

=====================

Q: How do we tell whether we are nested or not
during visits?
```

Without a pre/post-hook on parsing steps, I don't see how to address this in the parser at the moment.

Except, of course, via an AST walk.
How concerned are you about the cost of an AST walk? Honestly I can't imagine it will be too expensive.

### DIFF
--- a/src/parser/__tests__/parser-basic-test.js
+++ b/src/parser/__tests__/parser-basic-test.js
@@ -441,6 +441,57 @@ describe('basic', () => {
       },
       flags: ''
     });
+
+    expect(re('/(((a)b)c)/')).toEqual({
+      type: 'RegExp',
+      body: {
+        type: 'Group',
+        capturing: true,
+        number: 1,
+        expression: {
+          type: 'Alternative',
+          expressions: [
+            {
+              type: 'Group',
+              capturing: true,
+              number: 2,
+              expression: {
+                type: 'Alternative',
+                expressions: [
+                  {
+                    type: 'Group',
+                    capturing: true,
+                    number: 3,
+                    expression: {
+                      type: 'Char',
+                      value: 'a',
+                      kind: 'simple',
+                      symbol: 'a',
+                      codePoint: 97
+                    }
+                  },
+                  {
+                    type: 'Char',
+                    value: 'b',
+                    kind: 'simple',
+                    symbol: 'b',
+                    codePoint: 98
+                  }
+                ]
+              }
+            },
+            {
+              type: 'Char',
+              value: 'c',
+              kind: 'simple',
+              symbol: 'c',
+              codePoint: 99
+            }
+          ]
+        }
+      },
+      flags: ''
+    });
   });
 
   it('empty group', () => {
--- a/src/parser/__tests__/parser-basic-test.js
+++ b/src/parser/__tests__/parser-basic-test.js
@@ -381,7 +381,7 @@ describe('basic', () => {
     );
   });
 
-  it.only('named unicode name', () => {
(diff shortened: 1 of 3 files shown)


DmitrySoshnikov:
@davisjam, awesome, thanks; I'll try take a closer look at the PR soon.

davisjam:
>  I'm not sure how to construct the list of nodes in a way that enables the re-numbering we want, *without* traversing the full tree.

Suppose we did a single walk after parsing, to assign the "walk indices" of the nodes. Then each PostParsingPass could sort the nodes of interest by walk index before processing, providing a well-defined traversal order with O(number of nodes of interest) for each pass.

The walk index is the ordering needed for group renumbering so this would work for #164.
Do you see a need for other PostParsingPass's?

DmitrySoshnikov:
I've just recall that I had the https://github.com/DmitrySoshnikov/syntax/issues/48 opened for a while. This will allow analyzing a shifted token in the parser.

But now I'm thinking about adding a generic `onToken` event to the tokenizer, see the https://github.com/DmitrySoshnikov/syntax/issues/62. This allow subscribing, and checking if the parsed token is `L_PAREN`, i.e. `(`, and count.

```js
let totalGroups = 0;
let currentGroupNumber = 0;

yyparse.onParseBegin = (string, lexer) => {
  ...
  currentGroupNumber = 0;
  totalGroups = 0;
};

// Count left parens of capturing groups since
// In the /(((a)b)c)/, the "a" should be 3, "ab" - 2, "abc" - 1
yyparse.tokenizer.onToken = token => {
  if (token.type === 'L_PAREN' || token.type === 'NAMED_CAPTURE_GROUP') {
    currentGroupNumber++;
    totalGroups++;
  }
};
```

Then [when we have parsed the group](https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/parser/regexp.bnf#L777-L806), we do:

```js
currentGroupNumber--;

if (currentGroupNumber === 0) {
   currentGroupNumber = totalGroups;
}
```

So by the time it reaches the very nested group, the `currentGroupNumber` should be 3, then it decreases it, going up, until it reaches `0`. Then it resets to the total number, so in the:

```
/(((a)b)c)(d)/
``` 

the `"d"` will correctly be `4`.

This should allow us tracking the parens w/o doing post-processing.

davisjam:
Looks promising, I will take a look.

DmitrySoshnikov:
Cool! Make sure to use exactly `yyparse.onShift` hook, and not `yy.tokenizer.onToken`, since `onToken` happens prior the previous reduce, and it's hard to track the parens count.

DmitrySoshnikov:
OK, I've added the fix in https://github.com/DmitrySoshnikov/regexp-tree/commit/6709afd1192012e6caac60b4d4d8d38c646bb87a. @davisjam, please let me know if you see any issues. Closing this one.

davisjam:
Looks great. Sorry this fell off my radar.

PR #168, opened 2019-01-09
Change typescript definition file

Fazerty:
- export interfaces
- rename Regexp interface to AstRegexp to avoid confusion with Regexp
- export generate and toRegExp functions
- add the '?' kind to the SimpleQuantifier interface
- make the 'to' parameter optional in RangeQuantifier


DmitrySoshnikov:
Looks good, thanks!

PR #169, opened 2019-01-14
Fix: Change module name: regexp-tree/ast to regexp-tree_ast

Fazerty:


DmitrySoshnikov:
@Fazerty, can you add the description to this commit please, explaining why the change, etc?

Issue #179, opened 2019-05-19
Compat transpiler: Implement Unicode property escapes


DmitrySoshnikov:
Example:

```js
/\p{ASCII_Hex_Digit}/u
```

Compiles to:

```js
/[0-9A-Fa-f]/
```

See example in https://mothereff.in/regexpu (@mathiasbynens)

mathiasbynens:
Given the complexity and large amounts of Unicode data involved, would it make sense to re-use [the regexpu-core implementation](https://github.com/mathiasbynens/regexpu-core) for this?

I keep it up to date whenever Unicode updates, as well as [the Test262 tests](https://github.com/mathiasbynens/unicode-property-escapes-tests) and V8 implementation.

DmitrySoshnikov:
Perhaps -- depending on how large this dependency is. Currently I try to keep regexp-tree with no/minumum number of dependencies on external libraries. The tool has transformation API, so it's kinda weird to involve a separate transformation API/parser/etc. I'll need to take a closer look.

mathiasbynens:
I think it makes sense to have this transpiler be a separate regexp-tree plugin, in a separate repo + npm package (regardless of your decision to re-use regexpu-core or to do something else), for the reasons you describe. 

FYI, the raw data (based on the Unicode data) lives in https://github.com/mathiasbynens/regenerate-unicode-properties. If you decide not to reuse regexpu-core perhaps you could still use this data.

Issue #65, opened 2017-04-18
Compat-transpiler: implement `u` flag


DmitrySoshnikov:
The `u` flag should already be implemented by most of the engines, however, there are still old engines which do not implement it, so it'll still be nice to have this feature in compat-transpiler (ES2015 -> ES5).

There is an explicit implementation for the `u` flag in [regexpu](https://github.com/mathiasbynens/regexpu) package. Potentially can be ported to regexp-tree.

DmitrySoshnikov:
@mathiasbynens, you should know the topic, and details well, as an implementer of the `regexpu`, will be glad to get your reviews on this (in case you'd be willing to port it, I think you'd be the best owner for this feature).

mathiasbynens:
I’ve been meaning to port regexpu-core to a regexp-tree plugin and run some comparisons! Consider this assigned to me.

DmitrySoshnikov:
Sounds great, thanks! Let me know if you need any help, or any feature (and of course feel free to add any in case).

PR #182, opened 2019-06-27
compat-unicode-property-transform

JLHwung:
This is a WIP branch on addressing #179.

Above all, I would like to clarify the behavior of this plugin. It would only

- translate the `\p{...}` to the code points corresponding `CharacterClass`.

e.g. `\p{ASCII}` translated to `[\u0000-\u007F]`.
```js
{
      type: 'CharacterClass',
      expressions: [
        {
          type: 'ClassRange',
          from: {
            type: 'Char',
            value: '\\u{0000}',
            kind: 'unicode',
            symbol: '\u0000'
          },
          to: {
            type: 'Char',
            value: '\\u{007F},
            kind: 'unicode',
            symbol: '\u007F',
          }
        }
      ]
    }
```
- translate the `\P{...}` to the complement codepoints CharacterClass,
i.e. `\P{ASCII}` to `[\u0080-\u{10FFFF}]`.

- support `\u` flag only, which means output `\u{10FFFF}` instead of `\uDBFF\uDFFF` surrogate pair. Note: If we would really like to output surrogate pairs, I think there should be another compat transformer to transform code points outside BMP to surrogate pairs.

The transformed result would be far from optimized and the user is expected to leverage the builtin optimizer, i.e. `char-code-to-simple-char-transform`, to process the transformed result so that it can be readable in the eyes of human.

Compared to [`regexpu-core`](https://github.com/mathiasbynens/regexpu-core) which greatly optimizes the resulted transform. This transformer will only focus on transforming property to code points character class.

Another note is that the package separation (discussed [here](https://twitter.com/DmitrySoshnikov/status/1143580731987255296)) is not reflected yet. As introducing `lerna` would bring some devtools changes of this project, I would like to separate the package on phase 2 so that it would not mess up with feature branch.

As the feature is not yet implemented, tests fail as expected.

### DIFF
--- a/src/compat-transpiler/transforms/compat-unicode-property-transform.js
+++ b/src/compat-transpiler/transforms/compat-unicode-property-transform.js
@@ -10,14 +10,97 @@
  *
  */
 
+function getRegenerateSets(node) {
+  return [
+    'regenerate-unicode-properties',
+    node.binary ? 'Binary_Property' : node.canonicalName,
+    node.binary ? node.canonicalName : node.canonicalValue,
+  ].join('/');
+}
+
+function negateCodepoints(codepoints) {
+  return [0, ...codepoints, 0x110000];
+}
+
+function printCodepoint(codepoint) {
+  if (codepoint < 65536) {
+    return `\\u${codepoint
+      .toString(16)
+      .toUpperCase()
+      .padStart(4, '0')}`;
+  }
+  return `\\u{${codepoint.toString(16).toUpperCase()}}`;
+}
+
+function toCharLiteral(codepoint) {
+  return {
+    type: 'Char',
+    codePoint: codepoint,
+    kind: 'unicode',
+    value: printCodepoint(codepoint),
+    symbol: printCodepoint(codepoint),
+  };
+}
+
+function constructCharacterRange(node) {
+  //fixme: `regenerate.prototype.data` is private API.
+  let codepoints = require(getRegenerateSets(node)).data;
+  if (node.negative) {
+    codepoints = negateCodepoints(codepoints);
+  }
+  const CharacterRange = new Array(codepoints.length / 2);
+  for (let i = 0; i < codepoints.length; i += 2) {
+    if (codepoints[i + 1] - 1 === codepoints[i]) {
+      CharacterRange[i / 2] = toCharLiteral(codepoints[i]);
+    } else {
+      CharacterRange[i / 2] = {
+        type: 'ClassRange',
+        from: toCharLiteral(codepoints[i]),
+        to: toCharLiteral(codepoints[i + 1] - 1),
+      };
+    }
+  }
+
+  return CharacterRange;
+}
+
 module.exports = {
   shouldRun(ast) {
     return ast.flags.includes('u');
   },
-  CharacterClass() {
-    //todo: check unicode property inside character class, replace subarray of CharacterClass expression
+  CharacterClass(path) {
+    const {node} = path;
+    let newExpressions = [];
+    let lastUnicodePropertyIndex = 0;
+    for (let i = 0; i < node.expressions.length; i++) {
+      if (node.expressions[i].type === 'UnicodeProperty') {
+        newExpressions = newExpressions.concat(
+          node.expressions.slice(lastUnicodePropertyIndex + 1, i),
+          constructCharacterRange(node.expressions[i])
+        );
+        lastUnicodePropertyIndex = i;
+      }
+    }
+    if (newExpressions.length === 0) {
+      return;
+    }
+    newExpressions = newExpressions.concat(
+      node.expressions.slice(
+        lastUnicodePropertyIndex + 1,
+        node.expressions.length
+      )
+    );
+
+    path.replace({
+      type: 'CharacterClass',
+      expressions: newExpressions,
+    });
   },
-  UnicodeProperty() {
-    //todo: replace UnicodeProperty by CharacterClass
+  UnicodeProperty(path) {
+    const {node} = path;
+    path.replace({
+      type: 'CharacterClass',
+      expressions: constructCharacterRange(node),
+    });
   },
 };
--- a/src/compat-transpiler/transforms/compat-unicode-property-transform.js
+++ b/src/compat-transpiler/transforms/compat-unicode-property-transform.js
@@ -1,0 +1,23 @@
+/**
+ * The MIT License (MIT)
+ * Copyright (c) 2019-present Junliang Huang <jlhwung@gmail.com>
+ */
+
+'use strict';
+
+/**
+ * A regexp-tree plugin to translate `/\p{ASCII_Hex_Digit}/u` to `/[0-9A-Fa-f]/`.
+ *
+ */
+
+module.exports = {
+  shouldRun(ast) {
+    return ast.flags.includes('u');
+  },
+  CharacterClass() {
+    //todo: check unicode property inside character class, replace subarray of CharacterClass expression
+  },
+  UnicodeProperty() {
+    //todo: replace UnicodeProperty by CharacterClass
+  },
+};
--- a/src/compat-transpiler/transforms/compat-unicode-property-transform.js
+++ b/src/compat-transpiler/transforms/compat-unicode-property-transform.js
@@ -19,7 +19,18 @@ function getRegenerateSets(node) {
 }
 
 function negateCodepoints(codepoints) {
-  return [0, ...codepoints, 0x110000];
+  const newCodepoints = codepoints.slice();
+  if (newCodepoints[0] !== 0) {
+    newCodepoints.unshift(0);
+  } else {
+    newCodepoints.shift();
+  }
+  if (newCodepoints[newCodepoints.length - 1] !== 0x110000) {
+    newCodepoints.push(0x110000);
+  } else {
+    newCodepoints.pop();
+  }
+  return newCodepoints;
 }
 
 function printCodepoint(codepoint) {
--- a/src/compat-transpiler/transforms/compat-unicode-property-transform.js
+++ b/src/compat-transpiler/transforms/compat-unicode-property-transform.js
@@ -34,7 +34,7 @@ function negateCodepoints(codepoints) {
 }
 
 function printCodepoint(codepoint) {
-  if (codepoint < 65536) {
+  if (codepoint <= 0xffff) {
     return `\\u${codepoint
       .toString(16)
       .toUpperCase()
--- a/package-lock.json
+++ b/package-lock.json
@@ -4953,6 +4953,21 @@
       "integrity": "sha512-jVpo1GadrDAK59t/0jRx5VxYWQEDkkEKi6+HjE3joFVLfDOh9Xrdh0dF1eSq+BI/SwvTQ44gSscJ8N5zYL61sg==",
       "dev": true
     },
+    "regenerate-unicode-properties": {
+      "version": "8.1.0",
+      "resolved": "https://registry.npmjs.org/regenerate-unicode-properties/-/regenerate-unicode-properties-8.1.0.tgz",
+      "integrity": "sha512-LGZzkgtLY79GeXLm8Dp0BVLdQlWICzBnJz/ipWUgo59qBaZ+BHtq51P2q1uVZlppMuUAT37SDk39qUbjTWB7bA==",
+      "requires": {
+        "regenerate": "^1.4.0"
+      },
+      "dependencies": {
+        "regenerate": {
+          "version": "1.4.0",
+          "resolved": "https://registry.npmjs.org/regenerate/-/regenerate-1.4.0.tgz",
+          "integrity": "sha512-1G6jJVDWrt0rK99kBjvEtziZNCICAuvIPkSiUFIQxVP06RCVpq3dmDo2oi6ABpYaDYaTRr67BEhL8r1wgEZZKg=="
+        }
+      }
+    },
     "regenerator-runtime": {
       "version": "0.11.0",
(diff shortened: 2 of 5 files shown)


mathiasbynens:
I’m excited_ to see progress here. Thanks for kicking this off, @JLHwung!

> Compared to regexpu-core which greatly optimizes the resulted transform. This transformer will only focus on transforming property to code points character class.

FWIW, regexpu-core uses [regenerate](https://github.com/mathiasbynens/regenerate) to create optimal character classes and other patterns. Going forward, it would be nice to re-use that here instead of re-inventing the wheel.

DmitrySoshnikov:
@JLHwung, thanks, the plugin setup looks good. We'll defer merging until the actual implementation, when the tests will be passing (we can still work incrementally, transforming specific Unicode property groups in their own PRs if it's easier, instead of one giant PR).

> Note: If we would really like to output surrogate pairs, I think there should be another compat transformer to transform code points outside BMP to surrogate pairs

Yep, I agree, it should be another transform.

> The transformed result would be far from optimized and the user is expected to leverage the builtin optimizer, i.e. `char-code-to-simple-char-transform`, to process the transformed result so that it can be readable in the eyes of human

And yes, there should be a pipeline of transforms, which should run sequentially obtaining as optimized result as possible. (Later we can leave it up to a user which transform they want to apply. Probably eventually we can issue some warning/notice or an advice at transform time, something like: "you're using Unicode properties transform. Consider applying char code transform on top of it")

> As introducing `lerna` would bring some devtools changes of this project

Cool, I'll need to read more about `lerna`. Is this a standard/preferred way to manage packages in a mono-repo? Or can GitHub/NPM itself manage it correctly?

@mathiasbynens:

> regexpu-core uses regenerate to create optimal character classes and other patterns. Going forward, it would be nice to re-use that here instead of re-inventing the wheel.

Yeah, this was partially discussed in #65. As mentioned, in general I don't see a need to mix parser/transformer APIs (regexp-tree itself is positioned as a parser/transformer -- depending on another transformer would be weird). However, I agree we should reuse as much as possible in terms of data, or port `regenerate` module to a regexp-tree plugin (@JLHwung, this would be another useful module, which we can implement as a separate package).

JLHwung:
@DmitrySoshnikov I agree on transform-time user advise. Ideally every `compat` transform should be followed by some `optimize` transform if code legibility is major concern here.

> Is this a standard/preferred way to manage packages in a mono-repo?

AFAIK there is no standard way to manage packages, but lerna should be one of the preferred tools as it is used in Babel, Jest and many other open source JavaScript projects. Here is a [curated list](https://github.com/korfuri/awesome-monorepo) of mono repo management tools. 

> can GitHub/NPM itself manage it correctly

Both GitHub and NPM do not provide extra support on monorepo package management. Alternatively, yarn has rolled out [yarn workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/) years ago.

@mathiasbynens I would like to implement with the help of `regenerate-unicode-properties`. Here is some approaches:

a) Apply `regenerate.prototype.toRegExp` and translate `\p{...}` to a regular expression, since the transformer result should be a collection of AST nodes instead of a regex, we have to use `regexp-tree` to re-parse the translated regex into AST nodes.

b) Apply `regenerate.prototype.toArray` and translate `\p{...}` to an array of consisted code points. This approach will suffer great performance overhead if the code points are of thousands. Consider `\p{Unified_Ideograph}`, a `toArray` call will return an array of length 87887 but all we need is actually the range.

c) Access private property `data`, which is an array of code points range pairs. It seems convenient enough to have transformer build its own ClassRange from `data`. But we know it will be nightmare to use undocumented API from a maintainer's prospective.

> FWIW, regexpu-core uses regenerate to create optimal character classes and other patterns. Going forward, it would be nice to re-use that here instead of re-inventing the wheel.

`regexpu-core` is an awesome tool providing `\u` and `\p` flag supports. However, as an RegExp AST transformer, we will have to introduce an extra parse cycle to integrate with `regenerate` if we use `toRegex` API.

For this transformer, it relies only on the data located in PropList.txt. If `regenerate` can offer API access to the underlying `data`, i.e. `regenerate.prototype.codepointRanges`, it will definitely offer great convenience for AST transformer libraries.

JLHwung:
@DmitrySoshnikov I end up with private `data` API of `regenerate-unicode-properties` and it seems the plugin works well on common cases. It is not a giant PR and I would add more test cases.

src/compat-transpiler/transforms/compat-unicode-property-transform.js
```diff
@@ -0,0 +1,117 @@
...
+  if (newCodepoints[newCodepoints.length - 1] !== 0x110000) {
+    newCodepoints.push(0x110000);
+  } else {
+    newCodepoints.pop();
+  }
+  return newCodepoints;
+}
+
+function printCodepoint(codepoint) {
+  if (codepoint < 65536) {
```
mathiasbynens: More readable: `<= 0xFFFF`

(in general, use hexadecimal integer literals for code points)

mathiasbynens:
Instead of `regenerate.prototype.toRegExp`, it's probably better to use `.toString` (especially if you intend on reparsing).

JLHwung:
> Instead of regenerate.prototype.toRegExp, it's probably better to use .toString (especially if you intend on reparsing).

Well, I am actually trying to stay away from reparsing so I construct the `CharacterClass` from `regenerate.prototype.data`. It is straightforward to start here and craft a transformer focusing on property => codepoints only.

DmitrySoshnikov:
Cool, I'll take a detailed look into it.

Does this add dependency on `regenerate` (as an indirect dependency from `regenerate-unicode-properties`)? Do we really need `regenerate`'s API to get the data (like `addRange`)? I'm wondering if it's possible to extract data only to a repo.

To reiterate, I don't see a need/reason adding another parsing/transforming tool as a dependency to the parsing/transforming tool. If needed, we can port everything from regenerate here, or port regenrate to use regexp-tree's API/parser. If that would be a larger refactoring, and you still need to depend on regenerate, we can make this as just a separate repo with a dependency on regexp-tree and regenerate-unicode-properties.

mathiasbynens:
> To reiterate, I don't see a need/reason adding another parsing/transforming tool as a dependency to the parsing/transforming tool. If needed, we can port everything from regenerate here, or port regenrate to use regexp-tree's API/parser. If that would be a larger refactoring, and you still need to depend on regenerate, we can make this as just a separate repo with a dependency on regexp-tree and regenerate-unicode-properties.

I think there's some confusion here. [regenerate](https://github.com/mathiasbynens/regenerate) is not a parsing/transforming tool! It does _not_ involve parsing and it does not act on an AST. Can you elaborate on what you mean by "port regenerate to use regexp-tree's API/parser"? It doesn't make sense to me.

DmitrySoshnikov:
@mathiasbynens, oh, seems I confused `regenerate` (which is not a transformer) with `regexpu` (which is). What specifically does `regenerate` do? I'll take a closer look into it.

mathiasbynens:
The problem `regenerate` solves is: given a list of code points, produce an ASCII-safe JS-compatible regular expression pattern that only matches those code points.

By default it generates ES5-compatible output (i.e. no `u` flag) but optionally it can generate ES2015+ patterns that leverage the `u` flag. You can think of it as a code generator, but it seems like exactly the minimal component that you'd need to recreate here anyway (to turn the data into a regexp pattern).

JLHwung:
> It seems like exactly the minimal component that you'd need to recreate here anyway (to turn the data into a regexp pattern).

True and this is exactly what this PR is doing.

> I'm wondering if it's possible to extract data only to a repo

[Here](https://github.com/mathiasbynens/unicode-12.1.0) is a data-only repo. But this repo is lack of the data structure we need a.k.a code points ranges.

An ideal data repo for the purpose of this transfromer should
1) Consist of codepoints range for every `Binary_Property`, `General_Category`, `Script` and `Script_Extensions`
2) Equip with devtools to parse `PropList.txt` from UCD.
3) Release every year after Unicode releases a new version.

`1` and `2` are one-time efforts. But `3` is long-term commitment. I trust Mathias for his years efforts on Unicode tools. And if we setup a new data repo, it would be sort of tools fragmentation for communities.

@mathiasbynens Are there any chance the `unicode-x.y.z` package could include a _codepoint range_ format? If so we can instead depend on `unicode-x.y.z` and everyone would be happy.

mathiasbynens:
> > It seems like exactly the minimal component that you'd need to recreate here anyway (to turn the data into a regexp pattern).
> 
> True and this is exactly what this PR is doing.

Yes, that's what I meant when I was talking about "re-inventing the wheel" earlier. It's fine if that's what you decide to do --- but IMHO it just seems simpler to reuse `regenerate` which already does exactly what you want, and for which I already maintain/update the Unicode data for new Unicode releases through `regenerate-unicode-properties`.

> @mathiasbynens Are there any chance the unicode-x.y.z package could include a codepoint range format? If so we can instead depend on unicode-x.y.z and everyone would be happy.

This sounds super similar to what `regenerate-unicode-properties` already provides. What would be the benefit of a separate effort that does the same thing?

I'm not being facetious, but am honestly wondering/asking what would be the benefit of doing all this _without_ `regenerate` if you end up having to duplicate everything it's doing + duplicate the data packages anyhow.

DmitrySoshnikov:
Thanks for the clarifications guys.

Yeah, the data-only repo would be nice to have. However, if the functionality in `regenerate` (for ranges) is _really_ needed, we can depend on it. Otherwise, we can take only smaller portion of it, and add it directly to util functions here.

What I'm trying to make, is to keep `regexp-tree` itself almost free of dependencies (currently it's free). If some repo is needed only for some transform, we can/should make this transform as a separate repo as @JLHwung proposed using `lerna`.

JLHwung:
> it just seems simpler to reuse regenerate which already does exactly what you want

While I admit that in most cases `regenerate-unicode-properties` works good on property escape. It seems to me that it lacks negation `\P` API support, which I find it much easier to implement from the codepoint ranges. Anyway, I would be delighted to know if there is any way in `regenerate` to query the complement set.

> This sounds super similar to what regenerate-unicode-properties already provides. What would be the benefit of a separate effort that does the same thing?

AFAIK, the `regenerate-unicode-properties` offers `toRegExp`, `toString` and `toArray`. The first two is regexp related, the last one is listing all the code points inside the set. None of them can offer the most concise data structure behind `regenerate-unicode-properties` -- the codepoints ranges. For example, 
```js
require('regenerate-unicode-properties/Binary_Property/Unified_Ideograph').toArray()
// Array (87887) ...
```
while
```js
require('regenerate-unicode-properties/Binary_Property/Unified_Ideograph').data
// Array (28)
```
It is far more convenient and efficient to work on code points range. The code points range is library-agnostic but the data is deeply coupled to `regenerate` so we have to depend on `regenerate`. The `unicode-x.y.z` offers codepoints, RegExp and symbols but lack of codepoints range.

> I'm not being facetious, but am honestly wondering/asking what would be the benefit of doing all this without regenerate

I think it is just different approaches to tackle the same problem. An analogy here is `Babel + babili` v.s. `Terser`. `Terser` does a great job on minifying JavaScript source code, but Babel as a JavaScript Parser, can also develop its own minifier working on Babel's AST.

> If some repo is needed only for some transform, we can/should make this transform as a separate repo

@DmitrySoshnikov  I agree. As lerna requires all the packages be in subdirectories. The new directory structure will be like
```
packages
├── regexp-tree
├── regexp-tree-compat-unicode-property-transform
```
I suggest this change in another PR and we can hold this one until the new directory structure get stable and we have successful published a new package under `lerna`.

DmitrySoshnikov:
@JLHwung, thanks for the explanations.

> The `unicode-x.y.z` offers codepoints, RegExp and symbols but lack of codepoints range.

So is codepoints range the only thing which we need to use from `regenerate`? If eventually this (and later other) transform will be in a separate repo, I'm not worried directly depending on `regenerate`. Otherwise, we can just port needed small portion of functionality into a util method directly in `regexp-tree` (not really a "wheel reinvention", but an alternative dependency management). Both work for me.

> I suggest this change in another PR and we can hold this one until the new directory structure get stable and we have successful published a new package under `lerna`.

Yes, thanks, sounds good to me.

mathiasbynens:
> So is codepoints range the only thing which we need to use from regenerate?

Given a list of code point ranges and a list of lone code points (for those that don’t form a range), we need a way to produce a valid JavaScript regular expression pattern. That’s all that regenerate does.

DmitrySoshnikov:
@mathiasbynens, sounds good, thanks. Yes, once we have introduced packages management using `lerna` as @JLHwung mentioned, I'm fine depending on `regenerate` if it's needed in the unicode properties package.

JLHwung:
@DmitrySoshnikov Great. I have a heavy workload this week. 🤯Hopefully I would draft another PR in a couple of days.

If we can depend on `regenerate`, the plugin would be similar to 
[babel-plugin-transform-unicode-regex](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-unicode-regex/src/index.js#L14), which is a wrapper of `regenerate`.

mathiasbynens:
> If we can depend on `regenerate`, the plugin would be similar to
> [babel-plugin-transform-unicode-regex](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-unicode-regex/src/index.js#L14), which is a wrapper of `regenerate`.

I think you meant to say it’s a wrapper around `regexpu-core`, which uses `regenerate`.

regexpu-core is what turns `\p{ASCII}` into `/[\0-\x7F]/`. It’s what recognizes the JavaScript-supported Unicode property escapes and uses the Unicode code point data. It depends on regenerate to produce the final patterns, and for the Unicode data it consumes.

regenerate is a generic utility that turns “a set of code points” into a regular expression pattern that matches only those code points.

PR #184, opened 2019-08-22
Improve parser suppport for regexp group names

### LINKED ISSUE #144: Parser: Unicode group names and named group references
Currently parser doesn't consider group reference as a reference since doesn't normalize Unicode names to check:

```
/(?<\u{41}\u0042AB>\k<AB\u{41}\u0042>)/u
```

This should recognize `\k<AB\u{41}\u0042` as a group reference (`ABAB`).

rbuckton:
This PR improves the lexer token definition for `GROUP_NAME` to include valid unicode characters from the Unicode ID_Start and ID_Continue categories.

Fixes #144


scripts/generate-unicode-id-parts.js
```diff
@@ -0,0 +1,158 @@
...
+
+const MAX_UNICODE_NON_SURROGATE = 0xffff;
+const MAX_UNICODE_CODEPOINT = 0x10ffff;
+const isStart = c => /\p{ID_Start}/u.test(c);
+const isContinue = c => /\p{ID_Continue}/u.test(c);
+
+let idStartWriter = new Writer();
+let idContinueWriter = new Writer();
+
+for (let cp = 0; cp < MAX_UNICODE_CODEPOINT; cp++) {
```
mathiasbynens: This should have been `<=`

scripts/generate-unicode-id-parts.js
```diff
@@ -0,0 +1,158 @@
...
+  if (isContinue(ch)) {
+    idContinueWriter.push(cp);
+  }
+}
+
+console.log(`/**
+* Generated by scripts/generate-unicode-id-parts.js on node ${
+  process.version
+} with unicode ${process.versions.unicode}
+* based on http://www.unicode.org/reports/tr31/ and https://www.ecma-international.org/ecma-262/6.0/#sec-names-and-keywords
```
mathiasbynens: Consider pointing to the latest spec draft at https://tc39.es/ecma262/ instead of a versioned snapshot

scripts/generate-unicode-id-parts.js
rbuckton: Not sure that it matters that much, 0x10ffff is in the SPUA and likely won't ever be part of either unicode property.
```
> /[\p{ID_Start}\p{ID_Continue}]/u.test(String.fromCodePoint(0x10ffff))
false
```

mathiasbynens: I know, but if it was intentional, why stop at U+10FFFE?

```js
/[\p{ID_Start}\p{ID_Continue}]/u.test('\u{10fffe}');
// --> false
```

rbuckton: Not intentional, and I've pushed a fix. I was just commenting that it didn't affect the output.

scripts/generate-unicode-id-parts.js
```diff
@@ -0,0 +1,158 @@
...
+  }
+  if (isContinue(ch)) {
+    idContinueWriter.push(cp);
+  }
+}
+
+console.log(`/**
+* Generated by scripts/generate-unicode-id-parts.js on node ${
+  process.version
+} with unicode ${process.versions.unicode}
```
DmitrySoshnikov: So this depends on a Node version, and might be flaky/inconsistent? I'm wondering if it's possible just do a predictable subset, and update it periodically (even if manually).

Also, where/when will we run this script? Is it a part of the build script?

src/parser/regexp.bnf
```diff
@@ -431,19 +443,68 @@ function GroupRefOrDecChar(text, textLoc) {
...
+//  NonSurrogate ::
+//    Hex4Digits but only if the SV of Hex4Digits is not in the inclusive range 0xD800 to 0xDFFF    # [0-9a-ce-fA-CE-F][0-9a-fA-F]{3}|[dD][0-7][0-9a-fA-F]{2}
+//
+//  CodePoint ::
+//    HexDigits but only if MV of HexDigits ≤ 0x10FFFF                                              # 0*(?:[0-9a-fA-F]{1,5}|10[0-9a-fA-F]{4})
+//
+const uidRe = /\\u(?:([dD][89aAbB][0-9a-fA-F]{2})\\u([dD][c-fC-F][0-9a-fA-F]{2})|([dD][89aAbB][0-9a-fA-F]{2})|([dD][c-fC-F][0-9a-fA-F]{2})|([0-9a-ce-fA-CE-F][0-9a-fA-F]{3}|[dD][0-7][0-9a-fA-F]{2})|\{(0*(?:[0-9a-fA-F]{1,5}|10[0-9a-fA-F]{4}))\})/;
+
+function decodeUnicodeGroupName(name) {
+  return name.replace(new RegExp(uidRe, 'g'), function (_, leadSurrogate, trailSurrogate, leadSurrogateOnly, trailSurrogateOnly, nonSurrogate, codePoint) {
```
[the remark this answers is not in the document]
rbuckton: By adding a `/g` flag, the `lastIndex` is preserved between calls, even in the event of an unhandled exception. Rather than constantly mutating a value that is shared between multiple disparate calls, this ensures a fresh `RegExp` instance is created so that the `lastIndex` mutations remain local to the function.

scripts/generate-unicode-id-parts.js
rbuckton: It was run manually. Its unlikely this will change often, usually only if you choose to target a new version of NodeJS that itself includes a new version of the Unicode standard.

rbuckton:
The pre-commit I ran each file in the PR manually through prettier and didn't notice any changes. the pre-commit hook didn't trigger for me, possibly because I'm running on Windows?

src/parser/regexp.bnf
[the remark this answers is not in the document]
DmitrySoshnikov: Gotcha, OK.

rbuckton:
I just ran `prettier --write src/**/*.js` to see if I missed something, and it seems like the its the *untouched* parts of the repository actually have changes. Not sure if the fact I'm running on Windows might be the problem.

DmitrySoshnikov:
@rbuckton, do you use the same Prettier version installed from the `package.json`? (we'll update it later to the latest one). If it's a different version, it might change untouched changes. Let's change only your files, and we can update the whole repo to reformat with a new version of Prettier later.

rbuckton:
Yes, I ran prettier from `node_modules/.bin/prettier.cmd`. Also, it looks like the pre-commit hook did run, as the code *was* formatted from what I originally wrote.

rbuckton:
```
C:\dev\regexp-tree [fixUnicodeGroupNames ≡]> npm ls prettier
regexp-tree@0.1.11 C:\dev\regexp-tree
`-- prettier@1.17.1
```

rbuckton:
I started up bash for windows and ran the following:
```bash
rbuckton@rbuckton-book:/mnt/c/dev/regexp-tree$ jsfiles=$(git diff HEAD~2 --name-only --diff-filter=ACM "*.js" | tr '\n' ' ')
rbuckton@rbuckton-book:/mnt/c/dev/regexp-tree$ echo "$jsfiles" | xargs ./node_modules/.bin/prettier --write
scripts/generate-unicode-id-parts.js 172ms
src/compat-transpiler/transforms/compat-named-capturing-groups-transform.js 23ms
src/generator/__tests__/generator-basic-test.js 44ms
src/generator/index.js 46ms
src/parser/__tests__/parser-extended-test.js 55ms
src/parser/__tests__/parser-test262-test.js 53ms
rbuckton@rbuckton-book:/mnt/c/dev/regexp-tree$ git status
On branch fixUnicodeGroupNames
Your branch is up-to-date with 'origin/fixUnicodeGroupNames'.
nothing to commit, working directory clean
```

As far as prettier is concerned, there's nothing wrong. Is there a specific style issue I should address manually?

scripts/generate-unicode-id-parts.js
DmitrySoshnikov: OK, we can add it to the build script as one of the steps at some point.

DmitrySoshnikov:
OK, sounds good about Prettier. Later we'll need to setup Travis rule which would check against formatting with pre-defined locked version of the Prettier, so it's consistent.

Thanks, merging this one!

src/parser/regexp.bnf
```diff
@@ -22,7 +22,19 @@ U                       [0-9a-fA-F]
...
...ff]|\ud849[\udc00-\udfff]|\ud84a[\udc00-\udfff]|\ud84b[\udc00-\udfff]|\ud84c[\udc00-\udfff]|\ud84d[\udc00-\udfff]|\ud84e[\udc00-\udfff]|\ud84f[\udc00-\udfff]|\ud850[\udc00-\udfff]|\ud851[\udc00-\udfff]|\ud852[\udc00-\udfff]|\ud853[\udc00-\udfff]|\ud854[\udc00-\udfff]|\ud855[\udc00-\udfff]|\ud856[\udc00-\udfff]|\ud857[\udc00-\udfff]|\ud858[\udc00-\udfff]|\ud859[\udc00-\udfff]|\ud85a[\udc00-\udfff]|\ud85b[\udc00-\udfff]|\ud85c[\udc00-\udfff]|\ud85d[\udc00-\udfff]|\ud85e[\udc00-\udfff]|\ud85f[\udc00-\udfff]|\ud860[\udc00-\udfff]|\ud861[\udc00-\udfff]|\ud862[\udc00-\udfff]|\ud863[\udc00-\udfff]|\ud864[\udc00-\udfff]|\ud865[\udc00-\udfff]|\ud866[\udc00-\udfff]|\ud867[\udc00-\udfff]|\ud868[\udc00-\udfff]|\ud869[\udc00-\uded6\udf00-\udfff]|\ud86a[\udc00-\udfff]|\ud86b[\udc00-\udfff]|\ud86c[\udc00-\udfff]|\ud86d[\udc00-\udf34\udf40-\udfff]|\ud86e[\udc00-\udc1d\udc20-\udfff]|\ud86f[\udc00-\udfff]|\ud870[\udc00-\udfff]|\ud871[\udc00-\udfff]|\ud872[\udc00-\udfff]|\ud873[\udc00-\udea1\udeb0-\udfff]|\ud874[\udc00-\udfff]|\ud875[\udc00-\udfff]|\ud876[\udc00-\udfff]|\ud877[\udc00-\udfff]|\ud878[\udc00-\udfff]|\ud879[\udc00-\udfff]|\ud87a[\udc00-\udfe0]|\ud87e[\udc00-\ude1d]|\udb40[\udd00-\uddef])
```
DmitrySoshnikov: BTW, we'll need to keep an eye on whether this slows down the lexer significantly. This rule should run only with `u` flag, right?

rbuckton: No, the `U` flag only determines whether or not a `` `\u{` CodePoint `}` `` escape is supported and how `` `\u` Hex4Digits `` is handled. A `GroupName` always supports the Unicode `ID_Start` and `ID_Continue` properties: https://tc39.es/ecma262/#prod-RegExpIdentifierStart

```grammarkdown
GroupName[U] ::
  `<` RegExpIdentifierName[?U] `>`

RegExpIdentifierName[U] ::
  RegExpIdentifierStart[?U]
  RegExpIdentifierName[?U] RegExpIdentifierPart[?U]

RegExpIdentifierStart[U] ::
  UnicodeIDStart
  `$`
  `_`
  `\` RegExpUnicodeEscapeSequence[?U]

RegExpIdentifierPart[U] ::
  UnicodeIDContinue
  `$`
  `\` RegExpUnicodeEscapeSequence[?U]
  <ZWNJ>
  <ZWJ>

RegExpUnicodeEscapeSequence[U] ::
  [+U] `u` LeadSurrogate `\u` TrailSurrogate
  [+U] `u` LeadSurrogate
  [+U] `u` TrailSurrogate
  [+U] `u` NonSurrogate
  [~U] `u` Hex4Digits
  [+U] `u{` CodePoint `}`
```

As you can see, `UnicodeIDStart` and `UnicodeIDContinue` are referenced unconditionally.

scripts/generate-unicode-id-parts.js
mathiasbynens: I'll copy a comment I made on the TypeScript implementation here: https://github.com/microsoft/TypeScript/pull/32096#issuecomment-524203481

The build script could be simplified and made to not rely on the current Node.js/V8 + Unicode version:

```js
const ID_Start = require('unicode-12.1.0/Binary_Property/ID_Start/code-points.js');
const ID_Continue = require('unicode-12.1.0/Binary_Property/ID_Continue/code-points.js');

// ...then add the other needed characters, and write the two arrays to disk.
```

Links to how other tooling implements this:

- Esprima: https://github.com/jquery/esprima/blob/master/tools/generate-identifier-regex.js
- @babel/parser: https://github.com/babel/babel/blob/master/packages/babel-parser/scripts/generate-identifier-regex.js
- Acorn: https://github.com/acornjs/acorn/blob/master/bin/generate-identifier-regex.js
- JSHint: https://github.com/jshint/jshint/blob/master/scripts/generate-identifier-data.js
- XRegExp: https://github.com/slevithan/xregexp/blob/master/tools/scripts/utils.js

src/parser/generated/regexp-tree.js
```diff
@@ -1104,19 +1105,68 @@ function GroupRefOrDecChar(text, textLoc) {
 /**
  * Unicode names.
  */
-const uRe = /^\\u[0-9a-fA-F]{4}/;
-const ucpRe = /^\\u\{[0-9a-fA-F]{1,}\}/;
+const uReStart = /^\\u[0-9a-fA-F]{4}/; // only matches start of string
+const ucpReStart = /^\\u\{[0-9a-fA-F]{1,}\}/; // only matches start of string
+const ucpReAnywhere = /\\u\{[0-9a-fA-F]{1,}\}/; // matches anywhere in string
```
mathiasbynens: Is it a problem that this matches e.g. '\\\\u{2021}' (the even-odd problem)?

rbuckton: I'm not certain that would be the case. `\\` isn't otherwise allowed in a group name, which has already been parsed prior to this regexp being used.

PR #185, opened 2019-08-23
Improve types for regexp-tree

rbuckton:
Adds type information and exports for `transform`, `traverse`, and some other exported functions.


DmitrySoshnikov:
Sounds good, thanks! 

At some point we should probably migrate the whole code base to TypeScript directly.

PR #186, opened 2019-08-26
Enable all parser basic test

JLHwung:
Most tests of `parser basic` have been, I guess accidentally, deactivated since two years ago.

https://github.com/DmitrySoshnikov/regexp-tree/blob/eaf949cae817b6b264b0d2583e80ff12d0559fb0/src/parser/__tests__/parser-basic-test.js#L383-L390

In this PR we re-enable all parser basic test and remove invalid test cases. The following changes have been applied:

- Removed `non-named-backreferences with unicode` test, as it is considered a [syntax error](https://tc39.es/ecma262/#sec-patterns-static-semantics-early-errors) when a group specifier is not found in the enclosing patterns.

- Updated test artifacts related to `nameRaw` and `referenceRaw`.

- Apply babel transform only on publish step. Otherwise jest will call babel to transform the regexp literal in the unit test, i.e. `/\u{10FFFF}/` will be converted to `/\uDBFF\uDFFF/`.

### DIFF
--- a/src/parser/__tests__/parser-basic-test.js
+++ b/src/parser/__tests__/parser-basic-test.js
@@ -10,7 +10,6 @@ function re(regexp) {
 }
 
 describe('basic', () => {
-
   it('char', () => {
     expect(re(/a/)).toEqual({
       type: 'RegExp',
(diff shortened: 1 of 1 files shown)


JLHwung:
> probably we'll need to add a lint rule for .only at some point

FYI there is one: https://github.com/levibuzolic/eslint-plugin-no-only-tests

DmitrySoshnikov:
Yeah, I think we cad this one as a development dependency.

Issue #199, opened 2019-11-30
Option for disabling character class sorting in transformations


brettz9:
Hi,

While I expect this might not be trivial given that the sorting presumably helps in building ranges, it would be very useful if this resorting could be optionally disabled.

In the project, `eslint-plugin-unicorn`, `regexp-tree` is being used to otherwise very helpful effect in the rule [`regex-shorthand`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/regex-shorthand.md) (as per https://github.com/sindresorhus/eslint-plugin-unicorn/issues/453 ). This linting rule lints source files--files which are more likely than say distribution files--to call for preserving the original sorting order, whether for better readability or maintainability (e.g., `[aAbBcC]`) or for causing or necessitating additional modifications of regular expressions per the linting rule any time something might be manually added or changed in a character class to break the code point sorting pattern--especially given that users are not likely to always know or wish to spend time looking up code point values for the characters they are adding.

Thanks for your consideration!

voxpelli:
Adding parts of my comment from the other issue:

I agree that this is problematic.

One part of it is that different languages sort characters in different ways. Eg. in Swedish `åäö` is last in the alphabet and in that order and this rule eg. wants to resort `åä` to `äå`. To makes matter even more complicated, German sorts those letters after their dot-less equivalents, so `ä` after `a` and `ö` after `o` – so one universal true rule is not really possible when sorting purely on letters, unless one wants to enforce a "true" sorting upon everyone else.

DmitrySoshnikov:
Thanks, this makes sense. I have merged the PR, and also we can provide support for more granular options per-transform. E.g. for preserving the order in specific transform, instead of just fully disabling it. As a first step though, having the `blacklist` is a good option.

PR #200, opened 2019-12-03
Enable blacklisting of optimizations

voxpelli:
Makes it easier to disable some optimizations, as requested in eg. #199. Also adds some specific tests for one of the feature that's highlighted in #199 and is a partial fix for that issue.

### DIFF
--- a/src/optimizer/__tests__/optimizer-integration-test.js
+++ b/src/optimizer/__tests__/optimizer-integration-test.js
@@ -82,8 +82,26 @@ describe('optimizer-integration-test', () => {
     const original = '/(?:[a])/';
     const optimized = '/[a]/';
 
-    expect(optimizer.optimize(original, ['ungroup']).toString()).toBe(
-      optimized.toString()
-    );
+    expect(
+      optimizer.optimize(original, {whitelist: ['ungroup']}).toString()
+    ).toBe(optimized.toString());
+  });
+
+  it('sorts characters', () => {
+    const original = /[åä]/;
+    const optimized = /[äå]/;
+
+    expect(optimizer.optimize(original).toString()).toBe(optimized.toString());
+  });
+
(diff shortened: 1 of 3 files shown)


src/regexp-tree.js
```diff
@@ -123,8 +123,8 @@ const regexpTree = {
    *
    * @return TransformResult object
    */
-  optimize(regexp, whitelist) {
-    return optimizer.optimize(regexp, whitelist);
+  optimize(regexp, whitelist, {blacklist} = {}) {
```
DmitrySoshnikov: OK, so it's not a breaking change, and existing code which pass just the whitelist will continue to work 👍 Later we can add a deprecation notice, and eventually migrate to the single param with `{whitelist, blacklist}`.

src/optimizer/__tests__/optimizer-integration-test.js
```diff
@@ -82,8 +82,26 @@ describe('optimizer-integration-test', () => {
...
+    expect(optimizer.optimize(original).toString()).toBe(optimized.toString());
+  });
+
+  it('preserves character order when blacklisted charClassClassrangesMerge', () => {
+    const original = /[åä]/;
+    const optimized = /[åä]/;
+
+    expect(
+      optimizer
+        .optimize(original, {blacklist: ['charClassClassrangesMerge']})
```
DmitrySoshnikov: The `blacklist` is fine, although it disables the full transform. I think eventually we should also have `options` property, which would allow more granular configuring. Something like:

```js
{
  whitelist: ['charClassClassrangesMerge'],
  options: {
    charClassClassrangesMerge: {
      preserveCharOrder: true,
    }
  }
}
```

DmitrySoshnikov:
Thanks!

Merging this one, and if you'd like also to add support for per-transform `options`, would be great.

PR #205, opened 2020-03-02
Optimize API differences

brettz9:
Builds on #204 .

- Docs: Note difference between `regexp-tree` main `optimize` method and the optimizer method (esp. since `regexp-tree` README defers to this README)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "regexp-tree",
-  "version": "0.1.19",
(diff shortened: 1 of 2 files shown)


brettz9:
I've rebased. (I've also rebased #204 as the `package-lock.json` updated version is still not on `master`, though if you accept this PR, that PR will not be needed as this is built on top of that.)

DmitrySoshnikov:
Thanks!

PR #206, opened 2020-03-13
Fix `\s` rewrite

fisker:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Character_Classes#Types), `[\s]` should equivalent to `[ \f\n\r\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]`

### DIFF
--- a/src/optimizer/__tests__/optimizer-integration-test.js
+++ b/src/optimizer/__tests__/optimizer-integration-test.js
@@ -59,14 +59,14 @@ describe('optimizer-integration-test', () => {
   });
 
   it('whitespace', () => {
-    const original = /[ \n\r\t\f]+/;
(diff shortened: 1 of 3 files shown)


mathiasbynens:
regexpu-core produces the following:

```js
/[\t-\r \xA0\u1680\u2000-\u200A\u2028\u2029\u202F\u205F\u3000\uFEFF]/
```

See https://github.com/mathiasbynens/regexpu-core/blob/f68eb2aaa74adb491d3fb3a1991c224dd9c3397c/data/character-class-escape-sets.js#L12-L27.

DmitrySoshnikov:
Sounds good, thanks!

Should we have a separate transform then for ` \f\n\r\t\v` -> `\t-\r `?

PR #221, opened 2021-03-24
Type fixes

tjenkinson:
A few more type fixes.

Thanks again for the great lib!


DmitrySoshnikov:
Good catch, thanks! Might worth adding a test for this somehow, to avoid forgetting the new fields and their types.

PR #238, opened 2021-12-16
Fix sorting of caret

### LINKED ISSUE #237: Incorrectly rewrite `/[a^]/` to `/[^a]/`

fisker:
Fixes #237

### DIFF
--- a/src/optimizer/__tests__/optimizer-integration-test.js
+++ b/src/optimizer/__tests__/optimizer-integration-test.js
@@ -15,7 +15,7 @@ describe('optimizer-integration-test', () => {
     expect(optimizer.optimize(original).toString()).toBe(optimized.toString());
   });
 
-  it('preserves dash', () => {
+  it('preserves dash and caret', () => {
     let original = '/[^a-zа-яё -]+gi/';
     let optimized = '/[^ a-zа-яё-]+gi/';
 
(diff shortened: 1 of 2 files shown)


src/optimizer/__tests__/optimizer-integration-test.js
```diff
@@ -35,6 +35,32 @@ describe('optimizer-integration-test', () => {
...
+    expect(optimizer.optimize(original).toString()).toBe(optimized.toString());
+
+    original = /[^-]/;
+    optimized = /[^-]/;
+
+    expect(optimizer.optimize(original).toString()).toBe(optimized.toString());
+
+    // TODO: Fix this
+    original = /[-^]/;
+    optimized = /[^-]/;
```
fisker, line 61: This still broken, I'm not sure how to fix it. Maybe add some unsortable case?

DmitrySoshnikov: Yeah, this seems a regression, current version doesn't produce `/[^-]/` in this case. I think we should treat `^` as unsortable generically.

DmitrySoshnikov: Actually, maybe returning `return Number.MAX_SAFE_INTEGER` both for `-` and `^` would help?

src/optimizer/transforms/char-class-classranges-merge-transform.js
```diff
@@ -95,10 +95,13 @@ function sortCharClass(a, b) {
 function getSortValue(expression) {
   if (expression.type === 'Char') {
     if (expression.value === '-') {
-      return Infinity;
+      return Number.MAX_SAFE_INTEGER;
+    }
+    if (expression.value === '^') {
+      return Number.MAX_SAFE_INTEGER - 1;
```
DmitrySoshnikov, line 101: Should it also be `Number.MAX_SAFE_INTEGER`? In this case `/[-^]/` would stay the same.

src/optimizer/__tests__/optimizer-integration-test.js
fisker: No, `[-^]` `[^-]` has to be different, but current solution will break one of them, whatever we choose the order it. We have to find a different solution.

DmitrySoshnikov: I see – yes, then maybe putting them at the very end instead or adding extra slashes post-sorting?

PR #244, opened 2022-08-01
fix: add escaped special characters

### LINKED ISSUE #162: Regex with 'u' flag and escaped special characters can't be parsed
Hi,

Regex with unicode flag and some escaped special characters can't be parsed. For example the command `regexp-tree -e '/\./u'` returns "`invalid Unicode escape \.`". And also:

- `/\^/u`
- `/\$/u`
- `/\{/u`
- `/\}/u`
- `/\]/u`

The other special characters work (`*`, `+`, `?`, `(`, `)`, `|`, `[`, `\`).

regseb:
Fix #162.


DmitrySoshnikov:
Thanks for the fix. Did we check for potential side effects caused by this? Also, worth adding a test for this.

regseb:
It doesn't work because the list of special characters depends on the context.

- Normally there are: `*?+[()\|.^$`
- There are also `{}`, only they are in pairs and with a number inside.
- And in square brackets, there are only:
  - `]\`;
  - `-` if it is not in first or last position;
  - `^` if it is in first position.

And the list is different with unicode flag.

- Normally there are: `*?+[]()\|.^${}`
- And in square brackets, there are only:
  - `]\`;
  - `-` if it is not in first or last position;
  - `^` if it is in first position.

I don't know how to handle these cases with regular expressions.

---

```
Summary of all failing tests
 FAIL  src/parser/__tests__/parser-basic-test.js
  ● basic › character class with punctuation symbols

    SyntaxError: 
    
    /[!"#$%&'()*+,./:;<=>?@^_`{|}~-]/
                                ^
    Unexpected token: "}" at 1:28.

 FAIL  src/optimizer/transforms/__tests__/char-escape-unescape-transform-test.js
  ● e -> e › does not unescape { when looking like a quantifier

    SyntaxError: 
    
    /a\{3}/
         ^
    Unexpected token: "}" at 1:5.

  ● e -> e › does not unescape } when looking like a quantifier

    SyntaxError: 
    
    /a{3\}/
      ^
    Unexpected token: "{" at 1:2.
```

PR #254, opened 2023-04-09
Fix #227: treat "any single character" token as special in groupSingleCharsToCharClass

### LINKED ISSUE #227: optimizer incorrectly proposes rewrite .|[\r] as [.\r]
optimizer proposes replacing `.|[\r]` with `[.\r]` which is obviously incorrect as `.` plays different roles in these contexts

Maxim-Mazurok:
Fixes https://github.com/DmitrySoshnikov/regexp-tree/issues/227
and fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1231

### DIFF
--- a/src/optimizer/__tests__/optimizer-integration-test.js
+++ b/src/optimizer/__tests__/optimizer-integration-test.js
@@ -145,4 +145,14 @@ describe('optimizer-integration-test', () => {
         .toString()
     ).toBe(optimized.toString());
   });
+
+  [
+    /(?:.|\s)*/,
+    /^import ((?:.|\s)*?) from "(.*)";/gm,
+    /(.|[\n\r])/, // from https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1231#issue-870010385
+  ].map(regexp =>
+    it(`can recognise dot as "any single character" token in ${regexp}`, () => {
+      expect(optimizer.optimize(regexp).toString()).toBe(regexp.toString());
+    })
+  );
 });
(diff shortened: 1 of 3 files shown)


DmitrySoshnikov:
@Maxim-Mazurok thanks for the PR. Can we confirm all the tests pass, and do we need to handle other symbols?

Maxim-Mazurok:
> Can we confirm all the tests pass

Confirming that all tests are passing:
image

> do we need to handle other symbols?

I couldn't really think of anything that would have similar characteristics...

For example `/(\d|[\n\r])/gm` gets optimized into `/([\d\n\r])/gm` which is correct. Basically, we only need to care about symbols that mean different things depending on `[]` context.

ChatGPT suggested to check `^`, `-` and `\` which all seem to work fine. For example, `-` works fine because of this if I understand correctly: https://github.com/DmitrySoshnikov/regexp-tree/blob/091d4e9083abbc9f9573053d220c8a3db1fb3308/src/optimizer/transforms/char-class-classranges-merge-transform.js#L91-L102

So no, I don't think we need to handle other symbols.

DmitrySoshnikov:
Looks good, thanks for the contribution!

Maxim-Mazurok:
Hi @DmitrySoshnikov , looks like this PR was supposed to be included in v0.1.25 however I can't find expected "meta" word in https://www.unpkg.com/regexp-tree@0.1.25/dist/optimizer/transforms/group-single-chars-to-char-class.js and the issue is still reproducible in v0.1.25.

Perhaps it needs to be rebuilt and republished? Thank you!

DmitrySoshnikov:
@Maxim-Mazurok thanks for the heads up, yes, seems the build script is broken and didn't recompile the code (in fact, `npm run build` doesn't work anymore, I'll take a look).

DmitrySoshnikov:
@Maxim-Mazurok should be fixed in `v0.1.27`. Please let me know if you see any issues.

PR #264, opened 2024-06-23
fix: add support for `hasIndices` `/d` regex flag

aloisklink:
The `hasIndices` or `/d` regex flag is part of ES2022.

> The `d` flag indicates that the result of a regular expression match should contain the start and end indices of the substrings of each capture group. It does not change the regex's interpretation or matching behavior in any way, but only provides additional information in the matching result.

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/hasIndices


DmitrySoshnikov:
This looks good,  thanks for the contribution!

Issue #267, opened 2024-12-27
Incorrect quantifier merging in the optimizer


Aurele-Barriere:
The "quantifier merge transform" used by the optimizer is sometimes incorrect: a regex can be rewritten to another regex that is not equivalent.

- It it sometimes incorrect to rewrite `r{0,m}r{n}` into `r{n,m+n}`.

**Counter-example:**
`/(?:a|ab){0,1}(?:a|ab){1}/` is not equivalent to `/(?:a|ab){1,2}/`: they match different things on the string `"aba"`.

- For the same reason, it is incorrect to rewrite `r*r{n}` into `r{n,}`.

**Counter-example:**
`/(?:a|ab)*(?:a|ab){1}/` is not equivalent to `/(?:a|ab){1,}/`: they match different things on the string `"aba"`.

- It is sometimes incorrect to rewrite `r{0,n}?r{0,m}?` into `r{0,n+m}?`.

**Counter-example:**
`/(?:ab|aba){0,1}?(?:ab|aba){0,1}?[bc]/` is not equivalent to `/(?:ab|aba){0,2}?[bc]/`: they match different things on the string "ababc".

However, regexp-tree's optimizer module performs these transformations.

Thanks to Eugène Flesselle @eugeneflesselle for finding these counter-examples while working on proving the correctness of regex transformations.
We believe that these transformations are however correct when `r` is unambiguous, meaning that for each string, there is only one way of matching `/^r/`.
Maybe regexp-tree could check the ambiguity of the regex before applying these transformations.

DmitrySoshnikov:
@Aurele-Barriere thanks for reporting the issue (and @EugeneFlesselle for spotting it). Trying to keep balance between sophisticated and complicated optimizer, and handling all possible edge cases. How widespread is this construct? It'd be great to fix it anyways though. I'll appreciate a pull request on this if you reach it earlier.

PR #276, opened 2025-11-11
Optimizer: fix quantifier merge transform

Aurele-Barriere:
This PR addresses the issue https://github.com/DmitrySoshnikov/regexp-tree/issues/267 .

In some cases, the current quantifiers merge transform is incorrect.
This PR makes sure that the merging of quantifiers is only applied in cases that are known to be correct.

## When is quantifier merging correct?

We studied this problem recently in the following paper: https://arxiv.org/abs/2507.13091 .
Section 5 establishes conditions for which we proved that the optimization is correct.
Appendix B establishes counter-examples when these conditions aren't met.

From these results, we can establish 4 cases for when to perform the quantifier merge:
- When matching forward and the first quantifier is a forced quantifier (meaning that there are as many maximum iterations as minimum iterations, e.g. `a{4}`)
- When matching forward, both quantifiers are greedy, and the minimum iterations of the second quantifier is 0.
- When matching backward, and the second quantifier is a forced quantifier.
- When matching backwardn both quantifiers are greedy, and the minimum iterations of the first quantifier is 0.

Interestingly, this depends on the matching direction.
This means that we do not perform the same optimization when inside a lookbehind, or inside a lookahead (hence the `isForward` function).

## Correctness proofs

We wrote Rocq proofs, establishing that in these 4 cases it is correct to perform the optimization.

These proofs are available here: https://github.com/epfl-systemf/Linden/blob/f3c94efb9d5437073d65339c86d898ba669709e3/Rewriting/RegexpTree.v#L526-L640 .

## Work in Progress

This PR is still a draft, because this new version is still too restrictive: it should be possible to apply the optimization in more cases.
For instance, many tests in https://github.com/DmitrySoshnikov/regexp-tree/blob/master/src/optimizer/transforms/__tests__/quantifiers-merge-transform-test.js are now failing. And yet, in all of these cases, it should be correct to apply the optimization.

This is because the regex being quantified (only `a` in these tests) is non-ambiguous.
In such cases we believe that merging quantifiers is always correct.
My plan is to prove that this is correct, then implement a small non-ambiguity analysis and use it to perform the optimization more often.

### DIFF
--- a/src/optimizer/README.md
+++ b/src/optimizer/README.md
@@ -75,7 +75,7 @@ Here is the list of transforms supported by the Optimizer module.
 | charCodeToSimpleChar                                  | Don't use fancy char codes unless we have to                   | `\u0061` -> `a`               |
 | charCaseInsensitiveLowerCaseTransform                 | If regex is case insensitive, use lower-case everywhere        |  `/Aa/i` -> `/aa/i`           |
 | charClassRemoveDuplicates                             | Remove duplicates from char classes                            |  `[\d\d]` -> `[\d]`           |
-| quantifiersMerge                                      | Merge quantifiers where possible                               |  `a{1,2}a{2,3}` -> `a{3,5}`   |
+| quantifiersMerge                                      | Merge quantifiers where possible                               |  `a{1,2}a{0,3}` -> `a{1,5}`   |
 | quantifierRangeToSymbol                               | Reduce visual of quantifier ranges                             |  `a{1,}` -> `a+`              |
 | charClassClassrangesToChars                           | Replace char class ranges with chars                           |  `[a-a]` -> `[a]`             |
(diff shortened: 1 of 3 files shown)


DmitrySoshnikov:
@Aurele-Barriere thank you for following up on the issue! I like attention to this special case. Do the tests  pass? Do we need to add an extra test for this? Also, how does it affect performance? (should we gate it behind a flag or is the use case wide spread?).

Aurele-Barriere:
Hi, thanks for the answer!

No the tests are not passing currently, because they were checking that the optimization was being applied in cases where it's not applied anymore. I think that if I add this non-ambiguous analysis I was mentioning at the end of the PR, all existing tests should pass.

I would also like to add extra tests: for instance checking that the optimization is not applied in cases where we know it's incorrect. I'll try to add them.

For performance and how wide spread it is, I have no idea for now. I don't think it needs to be behind a flag, because I don't see how it would make performance worse, but I'd like to confirm that experimentally.

